PR #48, opened 2022-03-09
Allow for `package.function()` xref syntax.

FabienLelaquais:


jrobinAV (approved):
I pushed some break lines. It was faster to do it than to comment your PR.

PR #53, opened 2022-03-14
GUI User Manual reorganization

FabienLelaquais:
- Better home page for GUI User Manual and Visual Elements
- Dropped user_ prefix on User Manual pages for GUI


jrobinAV (approved):
Nice

PR #88, opened 2022-04-06
Fix Getting started bad link + update Getting Started navbar based on summary

florian-vuillemot:
image


FabienLelaquais (approved):
If it works it works

PR #96, opened 2022-04-07
Custom domain

florian-vuillemot:


FabienLelaquais (approved):
👍

CNAME
```diff
@@ -0,0 +1 @@
+docs.taipy.io
```
jrobinAV, line 1: break line at end of file

jrobinAV (approved):

CNAME
florian-vuillemot: This file is generated by Github. There is no break line, I just copy/past.

PR #114, opened 2022-04-15
Remove start Airflow documentation

khaitruong922:


florian-vuillemot (approved):
LGTM.
@toan-quach can you please try to run this doc on your tests before merging this PR?

docs/manuals/core/config/job-config.md
```diff
@@ -112,96 +112,16 @@ _airflow_ mode with the following config :
...
-    [JOB]
-    mode = "airflow"
-    airflow_folder = "my_custom_path"
-    ```
-
-!!! warning "Production setting"
-
-    Taipy starts Airflow in `standalone` mode. It is an Airflow development mode and not recommended for production.
-
 ## Using an external Airflow
```
jrobinAV: Do we need to keep this sub-section since we are always running with an external airflow?
```suggestion
```

Instead, I would add a small sentence at the beginning of the section 
`# Using Airflow (Enterprise version only)` to explain that the user need to have an external airflow running.

jrobinAV:
As a side remark, the branch name should be feature/... instead of doc/...

jrobinAV (approved):

FabienLelaquais (approved):

Issue #121, opened 2022-04-22
Concurrent deployed doc versions


FabienLelaquais:
**Description**

The Web site will hold parallel versions of the doc, seen as `docs.taipy.io/<branch_name>`.
Each release branch of taipy-doc, as well as `develop` will create a static site that can be deployed at the corresponding location.

**Expected change**

- mkdocs.yml / site_url must contain branch name (develop, release/*)
- getting-started/Readme files must use the branch name as well.
- Users can go to one doc version or another.
- [x] Update dev practices

PR #122, opened 2022-04-22
Concurrent deployed doc

florian-vuillemot:
#121

### DIFF
--- a/tools/setup_generation.py
+++ b/tools/setup_generation.py
@@ -28,6 +28,7 @@ import json
 import os
 import re
 import shutil
+import subprocess
 import warnings
 from datetime import datetime
 from inspect import isclass, isfunction, ismodule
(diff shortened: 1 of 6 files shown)


tools/setup_generation.py
```diff
@@ -513,8 +514,22 @@ def format_getting_started_navigation(filepath: str) -> str:
...
 
 
+
+
+# ------------------------------------------------------------------------
+# Step 4
+#   Generating site url
+# ------------------------------------------------------------------------
+print("Step 4/4: Generating the Site URL")
+current_branch_name = subprocess.run(['git', 'branch', '--show-current'], stdout=subprocess.PIPE).stdout.decode()[:-1]
```
FabienLelaquais: That is slightly dangerous.
The current branch (from taipy-doc) may be different than a real version number, like in a situation of a fix.
This could be better handled with a program argument. May be both.

FabienLelaquais (approved):
Looks really neat. Well done.

tools/setup_generation.py
florian-vuillemot: Putting the version in the mkdocs file is a good way to do it?
```
extra:
  version: 1.0
```

PR #127, opened 2022-05-03
Stop using GH-pages

florian-vuillemot:
We don't need to update the gh-page anymore because the 'develop' branch is available on 'readthedocs' (https://docs.taipy.io/en/develop/).

Github notifies Readthedocs on each pushed on the repository; RTD automatically pulls and deploys the latest version of the doc.

When merged we will be able to remove the _TAIPY_REPOS_ACCESS_TOKEN_ secrets.


FabienLelaquais (approved):
Even easier!

PR #134, opened 2022-05-05
Update Airflow config default

khaitruong922:
Update default value of `app_folder` and `airflow_dags_folder`

### DIFF
--- a/docs/manuals/core/config/job-config.md
+++ b/docs/manuals/core/config/job-config.md
@@ -219,7 +219,7 @@ Before executing a task, Airflow checks if its inputs are ready every 20 seconds
     airflow_sensor_poke_interval = "60:int"
     ```
 
-Because Airflow executes the application code in a different directory, you must make sure that all the files used by the application are accessible by the Airflow worker. We can configure the path to the application by specifying its absolute path or relative to the Airflow folder. The default path is "" (empty string), which implies that the absolute path of the running application will be used.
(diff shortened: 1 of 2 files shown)


florian-vuillemot (approved):
Top

docs/manuals/core/config/job-config.md
```diff
@@ -219,7 +219,7 @@ Before executing a task, Airflow checks if its inputs are ready every 20 seconds
     airflow_sensor_poke_interval = "60:int"
     ```
 
-Because Airflow executes the application code in a different directory, you must make sure that all the files used by the application are accessible by the Airflow worker. We can configure the path to the application by specifying its absolute path or relative to the Airflow folder. The default path is "" (empty string), which implies that the absolute path of the running application will be used.
+Because Airflow executes the application code in a different directory, you must make sure that all the files used by the application are accessible by the Airflow worker. We can configure the path to the application by specifying its absolute path or relative to the Airflow folder. The default value is "/opt/airflow/dags/app".
```
florian-vuillemot, line 222:
```suggestion
Because Airflow executes the application code in a different directory, you must make sure that all the files used by the application are accessible by Airflow workers. We can configure the path to the application by specifying its absolute path or relative to the Airflow folder. The default value is "/opt/airflow/dags/app".
```

jrobinAV (commented):
Approved !

PR #135, opened 2022-05-05
Update history and site versioning

florian-vuillemot:
Even if the _site_url_ is equal to _https://docs.taipy.io/en/release-1.0_, we can redirect from _https://docs.taipy.io/en/release-1.0_ to _https://docs.taipy.io/en/latest_ and thus deploy this version as "latest".
The only problem is that the redirect should follow the versions, and so we have to manage it.

Readthedocs redirection rule:
```
Redirection exacte
/en/release-1.0/$rest -> /en/latest/
```

### DIFF
--- a/docs/history.md
+++ b/docs/history.md
@@ -3,8 +3,29 @@ hide:
   - navigation
 ---
 
-# History
+# Change log
 
-## 1.0 (2022-04-07)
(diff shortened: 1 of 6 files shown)


jrobinAV (approved):

FabienLelaquais (approved):
👍

PR #173, opened 2022-07-13
Feature - REST doc generation

khaitruong922:


khaitruong922:
TODO: fetch taipy-rest source code and run the script to generate the json files

docs/manuals/rest/job.md
```diff
@@ -0,0 +1 @@
+!!swagger job.json!!
```
FabienLelaquais: I really don't think that these files (manuals/rest/<file>.md) should be commited any time.
They really look like they can be generated at the same time as their json content by tools/generate_rest.py...

tools/generate_rest.py
```diff
@@ -0,0 +1,33 @@
+import copy
```
FabienLelaquais, line 1: When and how is this script invoked?

Pipfile
```diff
@@ -40,6 +40,7 @@ simple-websocket = "==0.3"
 sqlalchemy = "==1.4"
 toml = "==0.10"
 tzlocal = "==3.0"
+mkdocs-render-swagger-plugin = "*"
```
FabienLelaquais: So I suspect that the Pipfile should be updated as well right?

.gitignore
```diff
@@ -121,3 +121,4 @@ docs/getting_started/index.md
 docs/getting_started/step_*/
 docs/getting_started/src/
 docs/getting_started/summary.txt
+docs/manuals/rest/*.json
```
FabienLelaquais, line 124: Probably the same rule for docs/manuals/rest/*.md

FabienLelaquais (changes requested):
That looks really, really promising!
Well done!!

Pipfile
khaitruong922: Yes because this is required for the rendering Swagger UI.

tools/generate_rest.py
khaitruong922: It should be invoked after taipy-rest and taipy-enterprise installed. How it is invoked should be similar to what we have done to taipy-core and taipy-gui ref manual, I am not really sure about this process.

docs/manuals/rest/job.md
khaitruong922: Sounds good to me.

.gitignore
khaitruong922: Ok but exclude the index.md

tools/generate_rest.py
FabienLelaquais: Ok so that would be part of the 'setup_generation' process.
How about adding a fourth step to the setup then?

.gitignore
FabienLelaquais: Right. Using the ! syntax, just like we already have already:
`docs/manuals/gui/viselements/*.md` except `index`.

mkdocs.yml_template
```diff
@@ -57,7 +57,7 @@ nav:
             - 'Job execution': 'manuals/core/entities/scheduling-and-job-execution.md'
     - 'Taipy REST':
         - manuals/rest/index.md
-        - 'Data node endpoints': 'manuals/rest/data-node.md'
+        - 'Data node endpoints': 'manuals/rest/datanode.md'
```
FabienLelaquais, line 60: These also could be generated. No big deal at this point.

FabienLelaquais (approved):
👍

jrobinAV (approved):

FabienLelaquais (approved):
I'll move things around when integrating.
I would like to see all this part of the Reference Manual...
Good job!

PR #184, opened 2022-07-26
feature/#40 added installation page

toan-quach:


docs/installation.md
```diff
@@ -0,0 +1,49 @@
...
+``` console
+$ pip install taipy
+```
+
+!!! info "This is the preferred method to install Taipy, as it will always install the most recent stable release."
+
+    If you don't have [pip][] installed, this [Python installation guide][]
+    can guide you through the process.
+
+    This method also applies for installing Taipy in an Anaconda environment.
```
jrobinAV: I would create 2 different sections. One for pip one for conda.

jrobinAV: I would also provide an example with conda explaining that pip is installed inside the conda virtual env: somewhere like "/anaconda/envs/my_venv_name/"

```
conda install pip
/anaconda/envs/my_venv_name/pip install taipy
```

docs/installation.md
```diff
@@ -0,0 +1,49 @@
+---
+hide:
+  - navigation
+---
```
jrobinAV, line 4: Why?

toan-quach: tbh, I'm not sure what it does, since I'm reusing the doc from the file installation.md in getting started, it just followed through with the move

docs/installation.md
toan-quach: Not sure I understand the part "/anaconda/envs/my_venv_name/", but yes I agree with having 2 different sections and also mentioning installing pip in the anaconda env

docs/installation.md
FabienLelaquais: You can use that when the page you're looking at is not in a context.
That would hide the navigation bar on the left.
It can be ok, it may not be.

docs/installation.md
```diff
@@ -0,0 +1,64 @@
...
+
+# Installation
+
+## Prerequisite
+
+To run Taipy, Python (version 3.8 or above) and [pip][] are required. If you don't have pip installed, this [Python installation guide][] can guide you through the process, or you can follow the [official installation page][] of pip.
```
FabienLelaquais: "[pip][]"?
"[Python installation guide][]", "[official installation page][]"??

"pip installed" -> "\`pip\` installed"

docs/installation.md
```diff
@@ -0,0 +1,64 @@
...
+
+You can also install pip in a [conda][] environment by using the following command:
```
FabienLelaquais: []?
as well as in the rest of this document.

docs/installation.md
```diff
@@ -0,0 +1,64 @@
...
+
+``` console
+$ pip install taipy
+```
+
+!!! info "This is the preferred method to install Taipy, as it will always install the most recent stable release."
+
+### Installing Taipy in a Conda environment
+
+Taipy can also be installed and used in a conda environment, to do so, run this command in your terminal:
```
FabienLelaquais: "in a conda environment, to do so, run"
->
"in a \`Conda\` environment: run"

docs/installation.md
FabienLelaquais: I mean I don't see the point of moving these Xrefs outside the text body.
This just makes is harder to maintain.

FabienLelaquais (approved):
Minor changes.

jrobinAV (approved):

docs/installation.md
toan-quach: @FabienLelaquais Ah I see, in this case, the installation page is under home page, so should we also expose the nav bar on the homepage?

docs/installation.md
toan-quach: agree, this also follows from the previous file. I will change it accordingly!

docs/installation.md
FabienLelaquais: Well that would make sense I guess...

docs/installation.md
```diff
@@ -7,9 +7,9 @@ hide:
 
 ## Prerequisite
 
-To run Taipy, Python (version 3.8 or above) and [pip][] are required. If you don't have pip installed, this [Python installation guide][] can guide you through the process, or you can follow the [official installation page][] of pip.
+To run Taipy, Python (version 3.8 or above) and [pip][https://pip.pypa.io] are required. If you don't have `pip` installed, this [Python installation guide][http://docs.python-guide.org/en/latest/starting/installation/] can guide you through the process, or you can follow the [official installation page][https://pip.pypa.io/en/latest/installation/] of `pip`.
```
FabienLelaquais: That I'm afraid would not be the Markdown syntax.
[text](link) would.

FabienLelaquais (changes requested):
Please check this link thing.

FabienLelaquais (approved):
👍

PR #187, opened 2022-07-28
Generate and integrate the REST API in the Reference Manual

FabienLelaquais:
- Generate REST API from scratch
- Remove Node.JS stuff
- GUI sources have moved to src/taipy


docs/manuals/reference_rest/index.md
```diff
@@ -0,0 +1,6 @@
+# Taipy's REST API
```
jrobinAV, line 1: This file is related to ref_man and not user_man. I would not put this file under `manuals` folder, but I would create another folder `reference`. 
The index.md file path would be `docs/reference/rest/index.md`.

Does it make sense to you?

jrobinAV (approved):
Great.

As soon as it is merged, I will work on improving the content itself.

Thx!

docs/manuals/reference_rest/index.md
FabienLelaquais: It does. This should be the 'reference_rest' directory.
Cannot use 'reference', which is 100 % generated...
Thanks!

FabienLelaquais: Oh. I was going to fix this.... but I don't need to.
We do agree: this file is specific to the Reference REST API Manual. And in my opinion, located at the right place.
Do you agree with that?

PR #194, opened 2022-08-25
feature/#276-synchronous-submit-method-for-async-dispatcher

toan-quach:
[link](https://app.zenhub.com/workspaces/taipy-613ef328affde6000ff77fb3/issues/avaiga/taipy-core/276)

### DIFF
--- a/docs/manuals/core/entities/scheduling-and-job-execution.md
+++ b/docs/manuals/core/entities/scheduling-and-job-execution.md
@@ -26,6 +26,20 @@ method triggers the submission of all the scenario's pipelines. Then each task o
         scenario.submit()
     ```
 
+By default, Taipy will asynchronously execute the jobs. If you want to wait until the submitted jobs are finished, you can use the parameter _wait_ and _timeout_:
+
+```python linenums="1"
+import taipy as tp
+import my_config
+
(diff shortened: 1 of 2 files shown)


docs/manuals/core/entities/scheduling-and-job-execution.md
```diff
@@ -78,6 +78,20 @@ task for execution.
...
+import my_config
+
+scenario = tp.create_scenario(my_config.monthly_scenario_cfg)
+task = scenario.predicting
+
+tp.submit(task, wait=True, timeout=3)
+```
+
+_timeout_ can be an integer or a float. By default, _wait_ is False and _timeout_ is None. If _wait_ is True and _timeout_ is not specified or None, there is no limit to the wait time. If _wait_ is True and _timeout_ is specified, taipy will wait until all the submitted jobs are finished or up to _timeout_ seconds.
+
```
jrobinAV: I believe the standard submit usage is the scenario submit. So I would put this text after the scenario submission and use a scenario as an example. Then for Pipeline and task, we can just add a small sentence saying that the `submit` method can also wait for the end of the jobs using wait and timeout parameters just like the scenario submit method.

jrobinAV (approved):
Good to go for me.

PR #201, opened 2022-09-08
Fix/#199 Propagate changes from `taipy-gui` and `taipy-enterprise`

trgiangdo:
https://app.zenhub.com/workspaces/engineering-team-6315bfd19a5b920013dabbaf/issues/avaiga/taipy-doc/199

I've also add a commit to fix `os.path.dirname` return empty string on Ubuntu machine. [Link to the solution and explanation.](https://stackoverflow.com/questions/7783308/os-path-dirname-file-returns-empty)


tools/setup_generation.py
```diff
@@ -40,7 +40,7 @@
 PACKAGE_GROUP = [ "taipy.config", "taipy.core", "taipy.gui", "taipy.rest", "taipy.auth", "taipy.enterprise" ]
 
 # Assuming that this script is located in <taipy-doc>/tools
-tools_dir = os.path.dirname(__file__).replace("\\", "/")
+tools_dir = os.path.dirname(os.path.abspath(__file__)).replace("\\", "/")
```
FabienLelaquais, line 43: What was the problem with that?

Pipfile
```diff
@@ -9,12 +9,13 @@ Flask-RESTful = "==0.3.9"
 apispec = {extras = ["yaml"], version = "==5.0"}
 apispec-webframeworks = "==0.5.2"
 bcrypt = "==3.2.2"
-flask = "==2.1"
+flask = "==2.2.2"
 flask-cors = "==3.0.10"
 flask-marshmallow = "==0.14"
 flask-socketio = "==5.1.1"
```
FabienLelaquais: Actually all this section should be changed as well:
```
flask = "==2.1"
flask-cors = "==3.0.10"
flask-marshmallow = "==0.14"
flask-socketio = "==5.1.1"
flask-talisman = "==1.0"
```
=>
```
flask = "==2.2.2"
flask-cors = "==3.0.10"
flask-marshmallow = "==0.14"
flask-socketio = "==5.3.0"
gevent = "==21.12.0"
gevent-websocket = "==0.10.1"
kthread = "==0.2.3"
```

FabienLelaquais (approved):
I have slightly different entries in Pipenv, as you'll see.
But well if it works, it works...
Thx!

tools/setup_generation.py
trgiangdo: On my Ubuntu WSL, the `os.path.dirname(__file__)` actually return an empty string, which is weird. I also did some searching and found others with similar issue on other Linux distro.

Pipfile
trgiangdo, line 15:
```suggestion
flask-socketio = "==5.3.0"
gevent = "==21.12.0"
gevent-websocket = "==0.10.1"
```
This should be ok, I suppose.
Does it necessary to move `gevent` and `gevent-websocket` below `flask-talisman` line?

trgiangdo:
```suggestion
flask-socketio = "==5.3.0"
gevent = "==21.12.0"
gevent-websocket = "==0.10.1"
```
This should be ok, I suppose.
Does it necessary to move `gevent` and `gevent-websocket` below `flask-talisman` line?

FabienLelaquais: The order is important only for us human beings: easier to locate an entry

jrobinAV (approved):
Looks good to me.
Thx

PR #217, opened 2022-09-15
Feature - SQL Table and SQL data node documentation

khaitruong922:


docs/manuals/core/config/data-node-config.md
```diff
@@ -22,7 +22,7 @@ More optional attributes are available on data nodes, including:
 -   _**storage_type**_ is an attribute that indicates the type of storage of the
     data node.<br/>
     The possible values are ["pickle"](#pickle) (**the default value**),
-    ["csv"](#csv), ["excel"](#excel), ["sql"](#sql), ["in_memory"](#in-memory), or
+    ["csv"](#csv), ["excel"](#excel), ["sql"](#sql), ["sql_table"](#sql_table), ["in_memory"](#in-memory), or
```
jrobinAV: JSON is missing. can you add it ?

docs/manuals/core/config/data-node-config.md
```diff
@@ -244,8 +244,7 @@ parameters can be provided.
 -   The _**db_engine**_ parameter represents the engine of the database.
 -   The _**read_query**_ parameter represents the SQL query that will be used by Taipy to read the data from the
     database.
--   The _**write_table**_ parameter represents the name of the table in the database that Taipy will be writing the
-    data to.
+-   The _**write_query_builder**_ parameter is the callback that takes in the data and return a list of queries that are executed when the data node is written into.
```
jrobinAV:
```suggestion
-   The _**write_query_builder**_ parameter is a callable function that takes the data as an input parameter and returns a list of SQL queries to be executed when the write data node method is called.
```

docs/manuals/core/config/data-node-config.md
```diff
@@ -223,16 +223,16 @@ the _exposed_type_. We also provide the list of specific sheets we want to use a
...
 !!! Important
 
     To be able to use a `SQLDataNode^` with Microsoft SQL Server you need to run internal dependencies with
     `pip install taipy[mssql]` and install your corresponding
     [Microsoft ODBC Driver for SQL Server](https://docs.microsoft.com/en-us/sql/connect/odbc/microsoft-odbc-driver-for-sql-server).
 
-A `SQLDataNode^` is a specific data node used to model Sql
-data. To add a new _sql_ data node configuration, the `Config.configure_sql_data_node()^` method can be used. In
+A `SQLDataNode^` is a specific data node used to model SQL
+data using queries. To add a new _sql_ data node configuration, the `Config.configure_sql_data_node()^` method can be used. In
```
jrobinAV:
```suggestion
An `SQLDataNode^` is a specific data node used to model data stored in an SQL database. To add a new _sql_ data node configuration, the `Config.configure_sql_data_node()^` method can be used. In
```

docs/manuals/core/config/data-node-config.md
```diff
@@ -255,26 +254,82 @@ parameters can be provided.
...
+!!! Important
+
+    To be able to use a `SQLTableDataNode^` with Microsoft SQL Server you need to run internal dependencies with
+    `pip install taipy[mssql]` and install your corresponding
+    [Microsoft ODBC Driver for SQL Server](https://docs.microsoft.com/en-us/sql/connect/odbc/microsoft-odbc-driver-for-sql-server).
+
+A `SQLTableDataNode^` is a specific data node used to model data from a SQL table. To add a new _sql_table_ data node configuration, the `Config.configure_sql_table_data_node()^` method can be used. In
+addition to the generic parameters described in the previous section
+[Data node configuration](data-node-config.md), multiple
+parameters can be provided.
```
jrobinAV:
```suggestion
An `SQLTableDataNode^` is a specific data node that models data stored in a single SQL table. To add a new _sql_table_ data node configuration, the `Config.configure_sql_table_data_node()^` method can be used. In
addition to the generic parameters described in the previous section
[Data node configuration](data-node-config.md), multiple
parameters can be provided.
```

docs/manuals/core/config/data-node-config.md
```diff
@@ -255,26 +254,82 @@ parameters can be provided.
...
-                                               write_table= "forecast_table")
+                                               write_query_builder= write_query_builder)
 ```
 
 In the previous example, we configure a _sql_ data node with the id "forecasts". Its scope is the
 default value `SCENARIO`. The database username is "admin", the user's password is "password", the database name
 is "taipy", and the database engine is `mssql` (short for Microsoft SQL). The read query will be "SELECT \* from
-forecast_table", and the table the data will be written to is "forecast_table".
+forecast_table".
+The write query builder in this example is a callback that takes in a dataframe and return a list of queries. The first query will delete all the data in the table "forecast_table" and the second query is a prepared statement that takes in two values, which is the data from the two columns "date" and "nb_sales" in the dataframe. Since this is a prepared statement, it must be passed as a tuple with the first element being the query and the second element being the data.
```
jrobinAV: Can you add a sentence to explain that the data parameter type must be the same as the return type of the function of the task config that has the data node as output?

docs/manuals/core/config/data-node-config.md
```diff
@@ -255,26 +254,82 @@ parameters can be provided.
...
+                                               db_username="admin",
+                                               db_password="password",
+                                               db_name="taipy",
+                                               db_engine="mssql",
+                                               table_name="forecast_table")
+```
+
+In the previous example, we configure a _sql_table_ data node with the id "forecasts". Its scope is the
+default value `SCENARIO`. The database username is "admin", the user's password is "password", the database name
+is "taipy", and the database engine is `mssql` (short for Microsoft SQL). The table name is "forecast_table". When the data node is read, it will read all the rows from the table "forecast_table" and when the data node is written, it will delete all the data in the table and insert the new data.
```
jrobinAV:
```suggestion
In the previous example, we configure a _sql_table_ data node with the id "forecasts". Its scope is the
default value `SCENARIO`. The database username is "admin", the user's password is "password", the database name
is "taipy", and the database engine is `mssql` (short for Microsoft SQL). The table name is "forecast_table". When the data node is read, it will read all the rows from the table "forecast_table," and when the data node is written, it will delete all the data in the table and insert the new data.
```

jrobinAV:
What about having the section sql_table before the section SQL? the SQL table case is more standard and common.

docs/manuals/core/config/data-node-config.md
khaitruong922: In Python we do not have static type in the parameter so there is not any restriction. It's up for the user to know what the return type of the task function is and code the write_query_builder accordingly.

jrobinAV: Yes of course, but still the data returned by the task must match with the way it is used within the query_builder. We must explain that so that the user understands it.
For instance, if the query_builder expects a data frame, the task must not return a list of integers.

jrobinAV (approved):

Issue #206, opened 2022-09-12
Scenario comparator documentation


jrobinAV:
Add documentation on user man about scenario comparators
- add a section on [scenario configs](https://docs.taipy.io/en/develop/manuals/core/config/scenario-config/) to explain how to configure a comparator
- add a section on [scenario management](https://docs.taipy.io/en/develop/manuals/core/entities/scenario-cycle-mgt/) to explain how to compare scenarios.

PR #221, opened 2022-09-20
Feature/#206 scenario comparator doc

toan-quach:
#206


docs/manuals/core/config/scenario-config.md
```diff
@@ -64,3 +64,15 @@ associated to a Cycle corresponding to their creation date. See documentation on
 [Scenario and cycle management](../entities/scenario-cycle-mgt.md).
 
 [:material-arrow-right: The next section introduces the global configuration](global-config.md).
+
+# Scenario configuration with comparators
```
FabienLelaquais: What about:
'Configuring scenario comparators'?

docs/manuals/core/config/scenario-config.md
```diff
@@ -64,3 +64,15 @@ associated to a Cycle corresponding to their creation date. See documentation on
...
+
+Here is a simple example. Assuming the pipeline configuration `pipeline_config` has already been created with `datanode_config` as one of the data node config in the pipeline config. The `compare_method` function is also predefined:
```
FabienLelaquais: Let us imagine a common situation where the pipeline configuration `pipeline_config` has been created with `datanode_config` as one of the data node configuration for this the pipeline configuration.
The function `compare_function` has also been defined:

docs/manuals/core/config/scenario-config.md
```diff
@@ -64,3 +64,15 @@ associated to a Cycle corresponding to their creation date. See documentation on
...
+
+```python
+from taipy import Config
+
+scenario_config = Config.configure_scenario("multiply_scenario", [pipeline_config], comparators={datanode_config.id: compare_method})
```
FabienLelaquais: compare_method -> compare_function

docs/manuals/core/config/scenario-config.md
```diff
@@ -64,3 +64,15 @@ associated to a Cycle corresponding to their creation date. See documentation on
...
+```
+
+In this example, we create a scenario configuration `ScenarioConfig^` from a pipeline configuration already defined. We also provide the scenario configuration with a dictionary of the data node configuration id that we want to compare and a predifined function that can be used to compare data nodes of the scenarios.
```
FabienLelaquais: We create the scenario configuration `scenario_config` using the indicated pipeline configuration. We use the `comparators` parameter to provide a dictionary indicating which data node need to be compared, with what function.

docs/manuals/core/entities/scenario-cycle-mgt.md
```diff
@@ -98,6 +98,23 @@ propagated to the nested pipelines, tasks, data nodes, and jobs if they are not
 
 If there is more than one scenario in the cycle, an error will be raised if you try to delete a primary scenario. You must promote another scenario as primary before deleting the primary one. On the other hand, if there is only one scenario in the cycle, the primary scenario can be deleted, which will also delete its cycle.
 
+# Compare scenarios
+
+You can compare two or more scenarios, created from the same scenario configuration, by using `taipy.compare_scenarios()` which takes two or more scenarios as the parameters. You can also provide the data node config id to specify which data nodes from the scenarios you want to compare through the parameter _**data_node_config_id**_.
```
FabienLelaquais: "which takes two or more scenarios as the parameters" 
->
". This function accepts two or more scenarios to be compared,"

FabienLelaquais: Remove the "through the parameter _**data_node_config_id**_" part

docs/manuals/core/entities/scenario-cycle-mgt.md
```diff
@@ -98,6 +98,23 @@ propagated to the nested pipelines, tasks, data nodes, and jobs if they are not
...
+from my_config import *
+
+previous_month_scenario = tp.create_scenario(monthly_scenario_cfg)
+current_month_scenario = tp.create_scenario(monthly_scenario_cfg)
+
+tp.compare(previous_month_scenario, current_month_scenario, "result_data_node")
+tp.compare(previous_month_scenario, current_month_scenario, "date_time_data_node")
+```
+
+!!! info "To see how you can define the function used to compare data nodes, please refer to the [Scenario Config](../config/scenario-config.md) page."
```
FabienLelaquais: ->
!!! info "Defining the comparison function"
To see how you can define the function used to compare data nodes, please refer to the [Scenario Config](../config/scenario-config.md) page.

FabienLelaquais (approved):
Changes that I think make things simpler.

docs/manuals/core/config/scenario-config.md
toan-quach: Sounds better!

docs/manuals/core/config/scenario-config.md
```diff
@@ -64,3 +64,16 @@ associated to a Cycle corresponding to their creation date. See documentation on
 [Scenario and cycle management](../entities/scenario-cycle-mgt.md).
 
 [:material-arrow-right: The next section introduces the global configuration](global-config.md).
+
+# Configuring scenario comparators
+
+Let us imagine a common situation where the pipeline configuration `pipeline_config` has been created with `datanode_config` as one of the data node configuration for this the pipeline configuration.
+The function `compare_function` has also been defined:
```
jrobinAV: I believe we should add more details on `compare_function`. Indeed, this `compare_function` must be able to accept as a parameter a list of data to compare. Each data from this list is an output of the data node read method, where the data node is instantiated from `datanode_config`.

Maybe we can take an example.

docs/manuals/core/entities/scenario-cycle-mgt.md
```diff
@@ -98,6 +98,24 @@ propagated to the nested pipelines, tasks, data nodes, and jobs if they are not
 
 If there is more than one scenario in the cycle, an error will be raised if you try to delete a primary scenario. You must promote another scenario as primary before deleting the primary one. On the other hand, if there is only one scenario in the cycle, the primary scenario can be deleted, which will also delete its cycle.
 
+# Compare scenarios
+
+You can compare two or more scenarios, created from the same scenario configuration, by using `taipy.compare_scenarios()`. This function accepts two or more scenarios to be compared, you can also provide the data node config id to specify which data nodes from the scenarios you want to compare.
+
+```python linenums="1"
+import taipy as tp
+from my_config import *
```
jrobinAV, line 108: On my_config.py, there is no comparator defined. Can you create a basic example in the my_config.py with 2 comparators (one for "result_data_node" and one for "date_time_data_node")?

Then we could display the output of both calls to `tp.compare()`.

It would be easier to understand for the user. What do you think ?

toan-quach: @jrobinAV I'll change the functions to comparing the `sales_predictions` data node as it's already in the my_config.py. We will create 2 scenario, 1 will have current month set as default datetime(2020, 1, 1), the other will have the current month written as datetime(2020, 2, 1), then we can do the comparison between the previous month scenario and this month scenario on the sales prediction made. What do you think?

docs/manuals/core/config/scenario-config.md
```diff
@@ -68,11 +68,17 @@ associated to a Cycle corresponding to their creation date. See documentation on
 # Configuring scenario comparators
 
 Let us imagine a common situation where the pipeline configuration `pipeline_config` has been created with `datanode_config` as one of the data node configuration for this the pipeline configuration.
-The function `compare_function` has also been defined:
+The function `compare_function` is also defined as followed:
 
 ```python
 from taipy import Config
 
+def compare_function(*data_node_results):
```
jrobinAV, line 82: What about making it completely explicit by naming the variable `data_node_read_results`? Is it a too long name?

docs/manuals/core/entities/scenario-cycle-mgt.md
jrobinAV: That sounds great!

docs/manuals/core/config/scenario-config.md
```diff
@@ -64,3 +64,22 @@ associated to a Cycle corresponding to their creation date. See documentation on
...
+The function `compare_function` is also defined as followed:
+
+```python
+from taipy import Config
+
+def compare_function(*data_node_results):
+  comparison_result = data_node_results[0]
+  for data_node in data_node_results[1:]:
+    comparison_result -= data_node
+  return comparison_result
```
jrobinAV: What about computing all possible comparisons? Something like:
```
# Calling compare_function(10, 13, 17, 9) returns the following dict
# {
# 0: {0: 0, 1: 3, 2: 7, 3: -1}, 
# 1: {0: -3, 1: 0, 2: 4, 3: -4}, 
# 2: {0: -7, 1: -4, 2: 0, 3: -8}, 
# 3: {0: 1, 1: 4, 2: 8, 3: 0}}
def compare_function(*data_node_results):
    comparison_result= {}
    current_result_index = 0
    for current_result in data_node_results:
        comparison_result[current_result_index]={}
        next_result_index = 0
        for next_result in data_node_results:
            print(f"comparing result {current_result_index} with result {next_result_index}")
            comparison_result[current_result_index][next_result_index] = next_result - current_result
            next_result_index += 1
        current_result_index += 1
    return comparison_result
```

May be it is too much. What do you think?

jrobinAV (approved):
I am good with the PR. I just made a non blocking suggestion on the compare_function example.

docs/manuals/core/config/scenario-config.md
toan-quach: @jrobinAV 
The datanode_config is very generic, so I went with something generic as well while keeping it simple so that it's easy to understand. But we can definitely make the compare function more complicated! Let me know what you think :D should we make it more complicated or the current version is good enough? I'll also put an example input and output at the beginning of the function like you did to make it extra clear 😄

docs/manuals/core/config/scenario-config.md
toan-quach: Sure! Makes it even clearer 😃

docs/manuals/core/config/scenario-config.md
toan-quach: Okay I'll take it back, it looks weird when saying "Calling the compare_function(10, 13, 17, 9) will return -29" 🤣

jrobinAV: Yes, indeed, that seems a bit strange to me.

jrobinAV (approved):

PR #222, opened 2022-09-21
BLOCKED until 2.0 is released - Feature/ Implement Mongo collection datanode

trgiangdo:
Documentation for PR https://github.com/Avaiga/taipy-core/pull/316

@toan-quach can you help me review and see if there is anywhere else in the doc that I need to add a note for Mongo datanode?


toan-quach (commented):
Some places I think you can mention MongoColectionDN:

1. data-node-config.md: 
- 1st part before Storage type section, where we talked in detail what each attribute the data_node_config needs. There's a detail mentioned of the `storage_type` and its possible values. Please add 'mongo' or whatever value we're looking for
- The `# Storage Type` section also mentions a little about the different values of the `storage_type`

2. The relnotes: might be too early to add to the release note, but it's for certain that we need to mention mongocollectionDN as a new feature in the relnote. You can add a new feature section for the new community version (I think we haven't decided on the next version number yet so best to leave it blank) and put mongocollectionDN as part of the new feature.

@jrobinAV @FabienLelaquais please add in your opinions of where else we should mention

jrobinAV:
Regarding the Release notes, I believe you are right @toan-quach. We need to wait for the release notes of 2.0 to be merged so we can initiate the 2.1 section. But still, it should be part of this PR.

jrobinAV (approved):

PR #236, opened 2022-10-12
Doc/ Add example to clarify JSON Encoder and Decoder

trgiangdo:
Documentation for https://github.com/Avaiga/taipy-core/pull/340

### DIFF
--- a/docs/manuals/core/config/data-node-config.md
+++ b/docs/manuals/core/config/data-node-config.md
@@ -436,8 +436,7 @@ hist_temp_cfg = Config.configure_json_data_node(
 In this example, we configure a JSON data node. The _id_ argument is "historical_temperature". Its _scope_ is
 `SCENARIO` (default value), and the path is the file hist_temp.json.
 
-Without specific _**encoder** and _**decoder**_ parameters, `hist_temp_cfg` will use the `DefaultJSONEncoder` and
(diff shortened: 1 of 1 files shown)


docs/manuals/core/config/data-node-config.md
```diff
@@ -414,16 +414,31 @@ In this example, we configure an in_memory data node with the id "date", the sco
...
+    default_path="path/hist_temp.json",
+)
+```
+
+In this example, we configure a JSON data node. The _id_ argument is "historical_temperature". Its _scope_ is
+`SCENARIO` (default value), and the path is the file hist_temp.json.
+
+Without specific _**encoder** and _**decoder**_ parameters, `hist_temp_cfg` will use the `DefaultJSONEncoder` and
+`DefaultJSONDecoder` provided by Taipy, which have the capability to encode and decode Python [`enum.Enum`](https://docs.python.org/3/library/enum.html), [`datetime.datetime`](https://docs.python.org/3/library/datetime.html#datetime-objects),
+and [dataclass](https://docs.python.org/3/library/dataclasses.html) object.
```
jrobinAV:
```suggestion
Without specific _**encoder** and _**decoder**_ parameters, `hist_temp_cfg` will use default encoder and decoder provided by Taipy, which have the capability to encode and decode Python [`enum.Enum`](https://docs.python.org/3/library/enum.html), [`datetime.datetime`](https://docs.python.org/3/library/datetime.html#datetime-objects),
and [dataclass](https://docs.python.org/3/library/dataclasses.html) object.
```

I removed the reference to the DefaultJSONEncoder class because it has been made private.

jrobinAV (approved):

PR #237, opened 2022-10-13
Split setup_generation file into multiple files

jrobinAV:
Proposal to start improving setup_generation script.

I did not change the logic at all. 
I just split the code into multiple files and functions, so that when we have a specific change in the code, we can just focus on the right file.

What we can do in the future: 
- Improve the code organization, having a dedicated folder to hold all the files
- Split each file into functions to improve code readability.

What do you think? I just spent 45 mins on it to initiate the first step of the code cleaning, so feel free to say that it is not enough or is not what you like. I won't be upset at all.


FabienLelaquais (approved):
👍

PR #243, opened 2022-10-19
Inject into config.py the methods to document in RefMan

jrobinAV:
To be merged in develop


FabienLelaquais (approved):
👌
With the two caveats: another step for config api integration, and caution with the code injection in the Config class.

PR #244, opened 2022-10-19
Inject into config.py the methods to document in RefMan

jrobinAV:
WARNING to be merged into release/2.0

### DIFF
--- a/tools/setup_generation.py
+++ b/tools/setup_generation.py
@@ -118,6 +118,7 @@ FORCE_PACKAGE = [
     ("taipy.core.data.operator.Operator", "taipy.core.data.operator"),
     ("taipy.core.data.operator.JoinOperator", "taipy.core.data.operator"),
     ("taipy.core.exceptions.exceptions.*", "taipy.core.exceptions"),
+    ("taipy.config.config.Config", "taipy.config"),
     ("taipy.config.checker.issue.Issue", "taipy.config"),
     ("taipy.config.checker.issue_collector.IssueCollector", "taipy.config"),
     ("taipy.rest.rest.Rest", "taipy.rest")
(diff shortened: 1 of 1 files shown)


tools/setup_generation.py
```diff
@@ -595,6 +596,56 @@ def generate_entries(entry_infos, package, type, package_output_file, in_group):
...
+from .common.scope import Scope
+from .common.frequency import Frequency
+from taipy.core.common.default_custom_document import DefaultCustomDocument
+from taipy.core.config.job_config import JobConfig
+from taipy.core.config.data_node_config import DataNodeConfig
+from taipy.core.config.task_config import TaskConfig
+from taipy.core.config.scenario_config import ScenarioConfig
+from taipy.core.config.pipeline_config import PipelineConfig\n"""
+    contents.insert(11, imports_to_inject)
+    contents.insert(len(contents) - 2, methods_to_inject)
```
FabienLelaquais, line 633: That is pretty dangerous, isn't it? If we change anything in the source file...
A safer (?) way could be to locate the class definition, skip the doc, then fill the body...

FabienLelaquais (approved):
Nice.
I think I would have created a new step for that... but hey

tools/setup_generation.py
jrobinAV: Yes, you are right.
I propose to do what you propose in the develop branch. This is for the 2.0 branch. In 2.0 branch, I believe config.py will not change.

jrobinAV:
> Nice. I think I would have created a new step for that... but hey

Good idea. In develop branch, the steps are well-designed. I will do it using a new step!

tools/setup_generation.py
FabienLelaquais: That makes a lot of sense. And if it does, we'll know soon enough.

Issue #254, opened 2022-10-27
BUG-When not Then


vhfgosselin:
**Bug description**
In the page https://docs.taipy.io/en/latest/manuals/gui/viselements/, we need to replace "then" by "when".

**Expected change**
"There are situations where you don't want a variable bound to a control value (such as the knob location of a slider) to be updated immediately **then** the user manipulates the control."

"There are situations where you don't want a variable bound to a control value (such as the knob location of a slider) to be updated immediately **when** the user manipulates the control."

**Screenshots**
When available and relevant, screenshots better help show the problem.

Issue #249, opened 2022-10-25
Rewrite tools/fetch_source_files.sh


FabienLelaquais:
**Description**

There are two problems with this script at of today:
- It is written using somewhat complex Bash constructs that some system don't support.
- We want to allow contributors to be able to build the doc without access to taipy-enterprise or taipy-auth.

**Expected change**

A Python version of this script would be portable, and can check for the access to enterprise & auth.

Issue #248, opened 2022-10-24
BUG-Wrong function name for long running callbacks 


FlorianJacta:
**Bug description**
 There is an error on the function name for the long running callback documentation in the [User Manual](https://docs.taipy.io/en/latest/manuals/gui/callbacks/#long-running-callbacks).

**Expected change**

The function name is `invoke_long_callback` and not `invoke_long_running`.

Issue #239, opened 2022-10-17
Outdated "Code organization" section in contributing doc


ghost:
**Description**
The "Code organization" in `CONTRIBUTING.md` and `docs/contributing/contributing.md` section is outdated.

**Expected change**
Update this section with appropriate informations.

PR #263, opened 2022-11-06
Rewrite fetch source files (#249)

FabienLelaquais:
Addresses #249, #248, #239 and #254.

Mainly, provides a Python version of tools/fetch_source_files.sh, for better portability and maintainability.


trgiangdo:
@FabienLelaquais We need an update on the [Installation page](https://github.com/Avaiga/taipy-doc/blob/develop/INSTALLATION.md).

Besides, what do you think about explicitly explain the arguments of `tools/fetch_source_files.py` on this page?
Like provide a list of `--no_pull`, `--check`, ... with short description on the installation page so that the user know the options, and they can run `--help` for more information.

FabienLelaquais:
> We need an update on the Installation page

Yes yes that's true. The different options are already documented in the sh and the py versions of the fetch script.
It would make a lot of sense to indicate them in the Installation md file as well, indeed.

At this stage, I have not changed the build process, which still relies on the Bash version of the script.
The plan is to move to he Python version, drop the sh version, and then update the Installation file, where your comment makes a lot of sense.

thx

tools/setup_generation/step_refman.py
```diff
@@ -397,34 +399,31 @@ def generate_entries(
...
 
         # Read config.py file
-        with open(path, 'r') as f:
+        with open(config_path, "r") as f:
             contents = f.readlines()
 
         # Inject imports and code
-        imports_to_inject = """from types import NoneType
-from typing import Any, Callable, List
+        imports_to_inject = """from typing import Any, Callable, List
```
jrobinAV, line 426: you removed NoneType? 
Is it on purpose?

jrobinAV (approved):
On the contributing.md File I have many remarks on the form and content. But I believe it is for another ticket. Right ?

tools/setup_generation/step_refman.py
FabienLelaquais: Well yes: it is not used... and it actually broke my setup at times...
Go figure.

PR #362, opened 2023-01-30
Propose core quick access buttons

jrobinAV:
Hello, 

I finally had time to propose quick access buttons for Core features.

Please tell me what you thnk.


ghost (commented):
I don't know what to think about this.
On one hand, I find it visually satisfying, but on the other hand, I think the pills should be used for more self-explainatory features.
From an accessibility perspective, having to hover a pill to see what it is about is not suitable as some important info is hidden by default.
At the risk of charging the page a little more, maybe I would have used a lighter version of the card component from quick access above...

jrobinAV:
I believed it was self-explanatory. If it s not, what about my new proposal?

ghost (changes requested):
Yeah, seems way better to me, indeed.
Maybe now you don't even need the `tp-tooltip` anymore, as it is mostly just redundant with the pill label.

jrobinAV:
Done

ghost (approved):
OK. Fine by me :)

jrobinAV:
@vhfgosselin @FlorianJacta Any feedback?

PR #381, opened 2023-02-08
Feature/contributors

jrobinAV:
Here is the current result:

image

- The list of Avaiga members is automatically built from github Avaiga organization. It only gets the accounts that accepted to set their Avaiga membership to public.
- The list of contributors is built from the join of all contributors.txt of all public repositories. We filter out the Avaiga members.
- Both list are randomly sorted.

Thx


mkdocs.yml_template
```diff
@@ -109,7 +109,6 @@ nav:
     - 'Contributing to Taipy': 'contributing/contributing.md'
     - 'Code of conduct': 'contributing/code_of_conduct.md'
   - 'Credits':
-    - 'Development lead': 'credits/development_lead.md'
```
FabienLelaquais, line 112: Why keep the 'Credits' title?
And not move this page under the 'Contributing' section. May be we can rename this...

tools/_setup_generation/setup.py
```diff
@@ -101,19 +102,25 @@ def exit(self, setup: Setup):
...
 from .step_getting_started import GettingStartedStep
-
-def run_setup(root_dir: str, steps: List[SetupStep] = [
-    VisElementsStep(),
-    RefManStep(),
-    RestRefManStep(),
-    GuiExtRefManStep(),
-    GettingStartedStep()
-    ]):
+from .step_contributor import ContributorStep
```
FabienLelaquais: I would have called this step 'contributors' unless you plan for no more than one.
(and ContributorsStep and so on)

tools/_setup_generation/step_contributor.py
```diff
@@ -0,0 +1,97 @@
...
+
+
+    def enter(self, setup: Setup):
+        self.PATH = os.path.join(setup.docs_dir, "credits", "contributors.md")
+
+    def get_id(self) -> str:
+        return "contributors"
+
+    def get_description(self) -> str:
+        return "Generating the contributors"
```
FabienLelaquais: the contributors list

(you're not generating contributors, are you?)

tools/_setup_generation/step_contributor.py
```diff
@@ -0,0 +1,97 @@
...
+
+    def setup(self, setup: Setup) -> None:
+        self.get_repo_urls()
+        self.get_avaiga_members()
+        self.get_contributors()
+        self.build_content((self.MEMBERS, "[AVAIGA_TEAM_MEMBERS]"), (self.CONTRIBUTORS, "[TAIPY_CONTRIBUTORS]"))
+
+    def get_repo_urls(self):
+        repos = requests.get(self.REPOS).json()
```
FabienLelaquais: Are you certain that this works for anonymous users?
I mean no access key, no nothing... and still works?

FabienLelaquais (approved):
It's a superb first proposal.
Love it.

A few minor remarks though.

mkdocs.yml_template
jrobinAV: Yep I agree.

I let's move "Contributors" page to the 2nd page of the "Contributing" doc set (after "Contributing to Taipy" but before "Code of conduct").

tools/_setup_generation/setup.py
jrobinAV: Yep!

tools/_setup_generation/step_contributor.py
jrobinAV: I would love to...

tools/_setup_generation/step_contributor.py
jrobinAV: I think so. I only get public information that you can get on GitHub directly.

PR #432, opened 2023-04-18
feature/#458 rename DefaultCustomDocument to MongoDefaultDocument

toan-quach:


jrobinAV (changes requested):
This is for 3.0 release. I propose to block it until taipy-core 2.3 is released.

Otherwise, it is fine.

jrobinAV (approved):

PR #451, opened 2023-05-19
feature/#569 remove scope.PIPELINE doc

toan-quach:


docs/manuals/core/concepts/scope.md
```diff
@@ -29,8 +26,8 @@ cycles.
 
     The first picture represents the basic case. We let Taipy apply the default scope (`Scope.SCENARIO`) for most
     data nodes. However, as you can see, the `sales predictions` data node is shared by the two pipelines
-    of my scenario. Consequently, its scope cannot be `Scope.PIPELINE`, but must be at least `Scope.SCENARIO`.
-    In this case, my various scenarios do not share any data nodes.
+    of my scenario. Consequently, its scope must be at least `Scope.SCENARIO`. In this case, my various scenarios
+    do not share any data nodes.
```
jrobinAV: We need to change that since Scope.SCENARIO is now the minimum level. So the sentence "Consequently, its scope must be at least `Scope.SCENARIO`." does not make sense anymore.

toan-quach: what about just `Consequently, its scope must be Scope.SCENARIO. In this case, my various scenarios`?

jrobinAV: Sounds good.

jrobinAV (approved):

Issue #442, opened 2023-05-02
Documentation on modifying change_delay globally


FlorianJacta:
**Description**
There needs to be documentation on how to modify the application's _change_delay_ value. It is possible to globally set it in the `<Gui>.run()`.

**Expected change**

A part of the documentation should explain how to change this value globally (_change_delay_).

Issue #439, opened 2023-04-29
GUI-Add documentation on configuration parameters


FabienLelaquais:
**Description**

The following configuration options are not listed in the Configuration section:
- [x] `allow_unsafe_werkzeug`
- [x] `async_mode`
- [x] `change_delay`
- [x] `chart_dark_template`
- [x] `extended_status`
- [x] `run_browser`
- [x] `run_in_thread`
- [x] `run_server`
- [x] `server_config`
- [x] `watermark`
- [x] `webapp_path`
- [x] `notebook_proxy`

It would be nice if they were...

**Expected change**

These are documented.

PR #455, opened 2023-05-24
Documentation for the Taipy GUI configuration settings (#439)

FabienLelaquais:
Addresses #442 and #439.


docs/manuals/run-deploy/deploy/linux/ubuntu.md
```diff
@@ -64,52 +64,54 @@ $python app.py
...
+[http://127.0.0.1:5000/](http://127.0.0.1:5000/).
 
 !!! note
     The message:
     ```
     WARNING: This is a development server. Do not use it in a production deployment.
     Use a production WSGI server instead.
     ```
-    Is provided by Flask because the way of exposing an application on the Internet and
+    Is issued by Flask because the way of exposing an application on the Internet and
```
FredLL-Avaiga: Should we add that this warning turns into an error when flask > 2.2.x ie is then blocking?

FredLL-Avaiga (approved):

dinhlongviolin1 (approved):
So far so good. 5 TODOs left

docs/manuals/run-deploy/deploy/linux/ubuntu.md
FabienLelaquais: Will do. Thx

FredLL-Avaiga (approved):

Issue #453, opened 2023-05-22
Update documentation on page scopes


FabienLelaquais:
**Description**

Page scoping has greatly been improved for 2.3.

**Expected change**

Doc is updated.

PR #464, opened 2023-06-04
Complement and clarify the page scope section

FabienLelaquais:
Addresses #453.


dinhlongviolin1 (approved):
LGTM. Just one more think is that should we add an example of importing a function from other page scopes to update a local variable

dinhlongviolin1 (approved):

Issue #458, opened 2023-05-30
Example of external page integration in parts


FabienLelaquais:
**Description**

Make sure the `part` block has an example showing the integration of an external page (ad area would be great),

**Expected change**

Doc is updated

FredLL-Avaiga:
cf. #785 and #786

PR #478, opened 2023-06-12
Add an example of a part embedding an external Web page

FabienLelaquais:
Addresses #458 and is a consequence of [taipy-gui#784](https://github.com/Avaiga/taipy-gui/issues/784) and [taipy-gui#785](https://github.com/Avaiga/taipy-gui/issues/785).


docs/manuals/gui/viselements/part.md_template
```diff
@@ -78,10 +79,10 @@ in the Stylekit.
 
 ### Showing and hiding controls
 
-!!! example "Page content"
+A `part` can be hidden by setting the [*render*](#p-render) property to False:
```
FredLL-Avaiga: technically it is not hidden, it's not there

FredLL-Avaiga (approved):

docs/manuals/gui/viselements/part.md_template
FabienLelaquais: I'll be more precise. Thanks.

PR #493, opened 2023-06-23
add entries in release notes for 2.3.1

jrobinAV:
release notes for 2.3.1

### DIFF
--- a/docs/relnotes.md
+++ b/docs/relnotes.md
@@ -59,6 +59,11 @@ Published on 2023-06.
   variables in complex dependencies situations.
 
 <h6 style="font-size: 1.2em"><strong><code>taipy-core</code></strong></h6>
+2.3.1
+- New API exposed at taipy package level to check if a scenario or a pipeline can be submitted (`is_submittable()^`)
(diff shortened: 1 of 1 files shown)


toan-quach (approved):
Well you took the job away from me but no complaints 🤣 LGTM

docs/relnotes.md
```diff
@@ -59,6 +59,12 @@ Published on 2023-06.
   variables in complex dependencies situations.
 
 <h6 style="font-size: 1.2em"><strong><code>taipy-core</code></strong></h6>
+2.3.1
```
FabienLelaquais, line 62: My shorter and less redundant proposal:

```
- New exposed functions to check access rights on entities:
   - `is_submittable()^` checks if a scenario or a pipeline can be submitted;
   - `is_promotable()^` checks if a scenario can be promoted to primary;
   - `is_deletable()^` checks if an entity can be deleted.
```

If you really want to say that they're accessible from the taipy package (which I think is useless):
```
These three functions are exported by the `taipy` package.
```

FabienLelaquais (approved):
I prefer my proposal. Surprise.

Issue #504, opened 2023-07-12
Document the Waterfall chart


FabienLelaquais:
**Description**

This addresses the documentation part of the [taipy-gui issue](https://github.com/Avaiga/taipy-gui/issues/839).

**Expected change**

The charts page has a new documented type.

PR #513, opened 2023-07-25
Waterfall charts doc

FabienLelaquais:
Addresses #504.


docs/manuals/gui/viselements/charts/waterfall.md
```diff
@@ -0,0 +1,245 @@
+# Waterfall charts
+
+Waterfall charts can display a running total as values are added or subtracted. They are used to
+rapidly understand how an initial value is affected by a series of positive or negative changes.
+
+A typical use of waterfall charts shows how a net value is reached through gains and losses
+over time.
+
+To create a waterfall chart in Taipy, you must set the [*type*](../chart.md#p-type) property
+of the chart control "waterfall".
```
FredLL-Avaiga: to "waterfall" ?

docs/manuals/gui/viselements/charts/waterfall.md
```diff
@@ -0,0 +1,245 @@
...
+want to display the running total: the cumulative difference between the actual score and the par
+for each hole. All those differences are stored in another array that we store in the *data*
+dictionary:
+
+```py
+data["Diff"] = [data["Score"][i]-data["Par"][i] for i in range(0,n_holes)] + [None]
+```
+
+For each hole, we compute the difference between the actual score and the par (a value of -2,
+for example, indicates ). We then set the entire array of differences, with an additional None
```
FredLL-Avaiga: indicates what ?

FredLL-Avaiga (approved):
👍

docs/manuals/gui/viselements/charts/waterfall.md
FabienLelaquais: Good spot!

docs/manuals/gui/viselements/charts/waterfall.md
FabienLelaquais: Who knows.

Issue #508, opened 2023-07-17
Change Pipeline/PipelineConfig documentation 


jrobinAV:
- [ ] Replace Pipeline by sequence
- [ ] Document the removal of taipy.create_pipeline()
- [ ] Document the rename of Config.configure_scenario_from_tasks() to Config.configure_scenario() to replace the old one.

trgiangdo:
TODO:
- [ ] Document the removal of `taipy.create_pipeline()`
- [ ] Document the rename of `Config.configure_scenario_from_tasks()` to `Config.configure_scenario()` to replace the old one.

PR #527, opened 2023-08-21
Doc/#508 - Remove deprecated taipy.create_pipeline() method

trgiangdo:
Preparation for https://github.com/Avaiga/taipy-doc/issues/508

### DIFF
--- a/docs/manuals/core/entities/pipeline-mgt.md
+++ b/docs/manuals/core/entities/pipeline-mgt.md
@@ -34,10 +34,10 @@ Taipy generates. A pipeline also holds various properties accessible as an attri
     from datetime import datetime
     import my_config
 
-    pipeline = tp.create_pipeline(my_config.sales_pipeline_cfg, name="Pipeline for sales prediction")
+    scenario = tp.create_scenario(my_config.monthly_scenario_cfg, name="Monthly scenario")
+    sales_pipeline = scenario.sales_pipeline
+    sales_pipeline.name = "Pipeline for sales prediction"
 
-    # The config_id is an attribute of the pipeline. Here it is "pipeline_configuration"
-    pipeline.config_id
     # There was no subscription, so subscribers is an empty list
     pipeline.subscribers # []
     # The properties dictionary equals {"name": "Pipeline for sales prediction"}. It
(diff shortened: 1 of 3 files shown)


docs/manuals/core/entities/pipeline-mgt.md
```diff
@@ -54,56 +54,26 @@ Taipy generates. A pipeline also holds various properties accessible as an attri
     current_month_data_node = pipeline.current_month
     ```
 
-# Get pipeline by id
+# Get a pipeline
 
-The method to access a pipeline is from its id by using the `taipy.get()^` method :
+There are two ways to retrieve a pipeline:
+
+The first method is by getting it from its parent scenario, using the pipeline name as an attribute.
```
toan-quach, line 60: should we also mention by using the `pipelines` attribute?

toan-quach (approved):
lgtm

docs/manuals/core/entities/pipeline-mgt.md
trgiangdo: `scenario.pipelines` would return a dictionary right? It doesn't return the entity directly, but it can work. 

I'm not sure though.

toan-quach: by that I meant `scenario.pipelines[pipeline_name]` :D

trgiangdo: Ah yes 😅

trgiangdo:
After thinking about it, I will not include the `configure_scenario()` and `configure_scenario_from_tasks()` in this PR, since it will include updating how to configure a scenario as well. This should be in a dedicated PR.

I propose keeping this PR to the removal of `create_pipeline()` only.

docs/manuals/core/entities/pipeline-mgt.md
```diff
@@ -34,10 +34,10 @@ Taipy generates. A pipeline also holds various properties accessible as an attri
     from datetime import datetime
     import my_config
 
-    pipeline = tp.create_pipeline(my_config.sales_pipeline_cfg, name="Pipeline for sales prediction")
+    scenario = tp.create_scenario(my_config.monthly_scenario_cfg, name="Monthly scenario")
+    sales_pipeline = scenario.sales_pipeline
+    sales_pipeline.name = "Pipeline for sales prediction"
```
jrobinAV, line 38: `scenario.sales_pipeline.name` always equals `sales_pipeline`.
Am I correct?

```suggestion
    sales_pipeline.name = "sales_pipeline"
```

jrobinAV (changes requested):
**In pipeline-mgt.md**
- In the pipeline attribute section, the `config_id` should be removed in line 17.
- The pipeline name is not valid. It should be a valid python identifier. No? Plus I believe it equals `sales_pipeline` in lines 39, 43, and 45.

**In scenario-cycle-mgt.md**
- We should add in `scenario-cycle-mgt.md` two entries for managing pipeline (add and remove) with a paragraph explaining that pipelines are initially created at the scenario creation from the scenario config. BUT we can also add new pipelines or remove default pipelines at run time. I don't know if it is already planned to be in another PR. If yes, please forget about the last point.

docs/manuals/core/entities/pipeline-mgt.md
trgiangdo: On this line, I'm assigning the name to `"sales_pipeline"` actually.

This is just to be consistent with the previous example.

trgiangdo:
@jrobinAV About adding/removing pipeline from scenario, that is not in the scope of this PR.
This is just to remove the deprecated `create_pipeline()` method.

Let's do that in other PRs.

docs/manuals/core/entities/pipeline-mgt.md
jrobinAV: Ah yes. thx.

jrobinAV (approved):

PR #544, opened 2023-09-02
Fix global navigation bar that disapeared

FabienLelaquais:
- Separate custom theme directory
- Separate User and Reference Manuals in "Manuals"
- Simplify indentation management when generating mkdocs.yml
- Change Twitter logo to X
- A few fixed xrefs


docs/manuals/index.md
```diff
@@ -0,0 +1,24 @@
+# About Taipy's Manuals
+
+The Taipy manuals are split into two categories:
```
jrobinAV, line 3: Minor suggestion: 
We are going to add tutorials, right? I would remove the number of categrories then.

```suggestion
The Taipy manuals are split into categories:
```

docs/manuals/usermans/index.md
```diff
@@ -0,0 +1,84 @@
...
+
+Taipy Core brings algorithm management to another level: algorithms are now connected to the end-user through
+user-defined scenarios, interactive data, smart job orchestration, etc.
+
+
+[:material-arrow-right: Definition of Taipy Core concepts](../core/concepts/index.md),
+
+[:material-arrow-right: Documentation on Taipy Core configuration](../core/config/index.md),
+
+[:material-arrow-right: Description of Taipy Core entities](../core/entities/index.md)
```
jrobinAV, line 71: Let's keep it for now since it is consistent with the actual user-man structure, but I will propose changes once we have designed a new structure.

jrobinAV (approved):

docs/manuals/index.md
FabienLelaquais: True. Because we don't have any yet, I would leave this as is.
When they come, this page will have to change anyway.

docs/manuals/usermans/index.md
FabienLelaquais: 👍

PR #549, opened 2023-09-06
Getting started - Install Taipy

erikfTW:
Please review, thanks.


FabienLelaquais (approved):
Looks fine to me.
Obviously, we'll have to move the Getting Started placeholder... when we have something to put there.

PR #572, opened 2023-09-22
Doc/Enterprise#219 - Scheduler

trgiangdo:


docs/relnotes.md
```diff
@@ -117,7 +117,8 @@ as well as additional features.
 
 ### New Features
 
-- ...
+- Taipy methods can be scheduled to run at a specific time using the new `taipy.Scheduler^` API.
```
jrobinAV:
```suggestion
- Python functions including scenario management methods can be scheduled to run at a specific time using the new `taipy.Scheduler^` API.
```

docs/manuals/core/scheduling/index.md
```diff
@@ -0,0 +1,278 @@
...
+For the scheduler to run the scheduled methods, first you need to start the scheduler using the `start()` method.
+
+```python
+import taipy as tp
+
+tp.Scheduler.start()
+```
+
+Once the scheduler is started, the scheduled methods will run in the background at the scheduled time.
+While the scheduler is running, you can run other methods without being blocked by the scheduler.
```
jrobinAV:
```suggestion
While the scheduler is running, you can schedule other methods without being blocked by the scheduler.
```

docs/manuals/core/scheduling/index.md
```diff
@@ -0,0 +1,278 @@
...
+# Submit a scenario every 1 week/month/year
+# Starting 1 week/month/year from now
+tp.Scheduler.every().weeks.do(tp.submit, monthly_scenario)
+tp.Scheduler.every().months.do(tp.submit, monthly_scenario)
+tp.Scheduler.every().years.do(tp.submit, monthly_scenario)
+
+tp.Scheduler.start()
+```
+
+In the above example, we schedule the `tp.submit(monthly_scenario)` method with different intervals.
```
jrobinAV, line 74:
```suggestion
In the above example, we schedule the `tp.submit(monthly_scenario)` method with different intervals. We chose to submit a scenario as an example, but any Python method can be scheduled.
```

docs/manuals/core/scheduling/index.md
```diff
@@ -0,0 +1,278 @@
...
+  `seconds`, `minutes`, `hours`, `days`, `weeks`, `months`, or `years`.
+- Finally, we provide the `do()` method with the method to schedule and its parameters.
+  In the example, we schedule the `tp.submit()` method, and provide `monthly_scenario` as the parameter.
+
+Once scheduled, the method will run in the background. The first run will be:
+
+- the number of interval from the scheduled time, or
+- at the start of the scheduler if the scheduler is started **after** the time that the method supposed to run.
+
+## Run a method at a specific time
```
jrobinAV:
```suggestion
## Run a method at a specific time and/or date
```

docs/manuals/core/scheduling/index.md
```diff
@@ -0,0 +1,278 @@
...
+from my_config import monthly_scenario_cfg
+
+monthly_scenario = tp.create_scenario(monthly_scenario_cfg)
+
+tp.Scheduler.every().second.do(tp.submit, monthly_scenario)
+
+tp.Scheduler.clear()
+```
+
+## Shortcut for scheduling Taipy methods
```
jrobinAV:
```suggestion
## Shortcut most used Taipy methods
```

jrobinAV (approved):

docs/manuals/core/scheduling/index.md
```diff
@@ -0,0 +1,278 @@
+# Taipy Scheduler
+
+!!! warning "Available in Taipy Enterprise edition"
+
+    This section is relevant only to the Enterprise edition of Taipy.
+
+This documentation focuses on providing necessary information to use the Taipy `Scheduler^`, and in particular
+the capabilities related to scenario management and job execution. It is assumed that the reader already knows the
```
erikfTW, line 8:
```suggestion
This documentation is primarily aimed at giving you the essential information needed to utilize the Taipy `Scheduler^`, with a specific emphasis on its capabilities regarding scenario management and job execution. It is assumed that the reader already knows the
```

docs/manuals/core/scheduling/index.md
```diff
@@ -0,0 +1,278 @@
...
+    ```python linenums="1"
+    {%
+    include-markdown "./code_example/my_config.py"
+    comments=false
+    %}
+    ```
+
+## Start/Stop the scheduler
+
+For the scheduler to run the scheduled methods, first you need to start the scheduler using the `start()` method.
```
erikfTW:
```suggestion
To make the scheduler execute the scheduled methods, you must first start the scheduler by using the `start()` method.
```

docs/manuals/core/scheduling/index.md
erikfTW:
```suggestion
Once the scheduler is started, the scheduled methods will run in the background at their specified times. 
While the scheduler is running, you can execute other methods without being held up by the scheduler.
```

docs/manuals/core/scheduling/index.md
```diff
@@ -0,0 +1,278 @@
...
+## Shortcut for scheduling Taipy methods
+
+Taipy `Scheduler^` provides shortcut for scheduling popular Taipy methods, including:
+
+- `taipy.create_scenario()`
+- `taipy.submit()`
+- A combination of `taipy.create_scenario()` and `taipy.submit()`
+
+Instead of calling the `do()` method and providing the Taipy method to schedule, you can call
+the following method directly from the scheduler.
```
erikfTW:
```suggestion
Instead of using the `do()` method and specifying the Taipy method to schedule, you can directly call the following method from the scheduler.
```

erikfTW (changes requested):

docs/manuals/core/scheduling/index.md
jrobinAV: PLease make the lines less than 100 char.

docs/manuals/core/scheduling/index.md
jrobinAV: I don't like the term `initiate` to talk about starting a service.
It s like a car you start it, you don't initiate it.

docs/manuals/core/scheduling/index.md
jrobinAV: Same remark here.
I don't like the term `activated` to talk about a service.
It s still like a car or an engine. It is not activated, it is running.
What do you think?

docs/manuals/core/scheduling/index.md
trgiangdo: I feel like the sentence is using too many advance words.

Tbh, I don't know what "with a specific emphasis on its capabilities regarding scenario management and job execution" means without re-reading it 2-3 times.

docs/manuals/core/scheduling/index.md
trgiangdo:
```suggestion
In the above example, we schedule the `tp.submit(monthly_scenario)` method with different intervals.
Notice that we chose to submit a scenario as an example, but any Python method can be scheduled.
```

docs/manuals/core/scheduling/index.md
trgiangdo: How about
```suggestion
## Shortcut for scheduling most used Taipy methods
```

docs/manuals/core/scheduling/index.md
erikfTW: OK maybe let's try this less advanced wording:
This documentation is centered around providing essential information for using the Taipy `Scheduler^`, 
particularly focusing on the capabilities related to scenario management and job execution. 
It is assumed that the reader already knows the

docs/manuals/core/scheduling/index.md
erikfTW: OK let's keep "start" then :) But "initiate" is a common word used in IT writing when you want to start/begin a process.

docs/manuals/core/scheduling/index.md
erikfTW: Yup, here I agree, that "activated" does not reflect directly that we want to say the scheduler is running. I have updated the commit, thanks.

docs/manuals/core/scheduling/index.md
trgiangdo: this is reasonable IMO. What do you think @jrobinAV?

docs/manuals/core/scheduling/index.md
jrobinAV: OK. My vocabulary is not so good. Let's use the most common word, whatever it is.

erikfTW: But both ways are OK from the English point of view, so we can keep "start" as well. The "start" wording is used quite often in other topics/sections as well, so to be consistent.

docs/manuals/core/scheduling/index.md
```diff
@@ -0,0 +1,278 @@
...
+For the scheduler to run the scheduled methods, first you need to start the scheduler using the `start()` method.
+
+```python
+import taipy as tp
+
+tp.Scheduler.start()
+```
+
+Once the scheduler is started, the scheduled methods will run in the background at the scheduled time.
+While the scheduler is running, you can schedule other methods without being blocked by the scheduler.
```
trgiangdo:
```suggestion
Once the scheduler is started, the scheduled methods will run in the background at their specified times. 
While the scheduler is running, you can execute other methods without being blocked by the scheduler.
```

erikfTW (approved):

PR #595, opened 2023-10-03
Fix/redesign regression

ghost:
Fix header and sidenav regression due to package upgrade and overrides misplacement.


FabienLelaquais (approved):
Sure thing

jrobinAV (approved):

PR #605, opened 2023-10-06
Reformulate the demo

RymMichaut:
Refomulating to make it more factual and less verbose


jrobinAV (approved):
Neat.
TODO JRM: I need to replace image tags on another PR.

PR #644, opened 2023-10-10
fix(header): Bring back dynamic site name and page title

ghost:
Bring dynamic site name and page title back in the header in order to keep context when scrolling content past h1.


FabienLelaquais (approved):
Whao that was huge!

PR #661, opened 2023-10-17
fix(style): decrease reference font-size on mobile

ghost:
Decrease reference font-size to 16px on mobile because, if 20px was good for desktop, it was way too big for mobile devices.

### DIFF
--- a/taipy_theme/assets/stylesheets/avaiga.css
+++ b/taipy_theme/assets/stylesheets/avaiga.css
@@ -81,6 +81,12 @@
   --md-hue: 270; 
 }
 
+@media screen and (max-width: 59.9375em){
+  html{
+    font-size: 100%;
+  }
+}
+
 .md-grid{
   max-width: var(--md-grid-max-width);
 }


taipy_theme/assets/stylesheets/avaiga.css
```diff
@@ -81,6 +81,12 @@
   --md-hue: 270; 
 }
 
+@media screen and (max-width: 59.9375em){
```
FabienLelaquais, line 84: 59.9375, seriously?

FabienLelaquais (approved):
👍

taipy_theme/assets/stylesheets/avaiga.css
ghost: Yeah. I dont like it either but that's a 959px on a 16px base, and that's the way it is done in Material Mkdocs CSS... 😔🔫
I personnaly prefer `@media screen and (max-width: calc(60em - 1px))` but hey... Code consistency.

Issue #663, opened 2023-10-17
Improvements & suggestions for clarity and consistency


ghost:
I went through the documentation pages and noticed some stuff that could be improved.
I reported it all [here](https://docs.google.com/document/d/1sqdVq7CpEnJOEDiM50aiaNYuNa_HWFdPLu922CASIpw/edit?usp=sharing).

jrobinAV:
Check what is still relevant.

jrobinAV:
Split issue into multiple tickets if necessary.

FlorianJacta:
Issues:
https://github.com/Avaiga/taipy-doc/issues/895
https://github.com/Avaiga/taipy-doc/issues/896
https://github.com/Avaiga/taipy-doc/issues/897
https://github.com/Avaiga/taipy-doc/issues/898
https://github.com/Avaiga/taipy-doc/issues/899

PR #665, opened 2023-10-17
feat(demos): make demo links look like buttons.

ghost:
Addresses #663 issue's point about demo buttons

image


FabienLelaquais (approved):
Love it

Issue #533, opened 2023-08-28
Document the Python API for page building in Taipy GUI


FabienLelaquais:
**Description**

A new way for building Taipy GUI pages has been created. This needs documentation and examples.

See [gui#898](https://github.com/Avaiga/taipy-gui/issues/898) for the implementation part,

**Expected change**

The page builder API has a dedicated User Manunal section.

PR #668, opened 2023-10-18
Page Builder API doc

FabienLelaquais:
Addresses #533.


FredLL-Avaiga (approved):

docs/manuals/gui/page_builder.md
```diff
@@ -0,0 +1,202 @@
...
+import src.taipy.gui.builder as tgb
+
+
+with tgb.Page() as page:
+    # add your visual elements here
+
+Gui(page).run()
+```
+
+The *page* object, after it has been added the required elements, is added to the `Gui` instance
```
dinhlongviolin1: `after it has been added the required elements,` this is not clear to me

docs/manuals/gui/page_builder.md
```diff
@@ -0,0 +1,202 @@
...
+
+The impact of this change is reflected in the page:
+
+<figure>
+    <img src="../tgb-3-d.png" class="visible-dark" />
+    <img src="../tgb-3-l.png" class="visible-light"/>
+    <figcaption>Styling HTML</figcaption>
+</figure>
+
+You can see that, compared to the previous example, the label uses a bold font weight.
```
dinhlongviolin1: After this, we can also state that the second parameter ("User Information") of an html class is optional and can be omitted

docs/manuals/gui/page_builder.md
```diff
@@ -0,0 +1,202 @@
...
+```py
+def my_button_pressed(state, id):
+    # Do something...
+    pass
+
+with tgb.Page() as page:
+    tgb.button("Press me", on_action=my_button_pressed)
+```
+
+This would have the same behavior as in the case where you would have used the function name.
```
dinhlongviolin1, line 215: We might want to address these points:
- Directly set a dictionary or array to properties, no need to map it through variable string 
```
tgb.chart(
            data="{csvdata}",
            x="Day",
            selected_color="green",
            y=["Daily hospital occupancy", "Daily hospital occupancy"],
            label=["Entity", "Code"],
            mode=[None, "markers"],
            color=[None, "red"],
            type=(None, "scatter"),
            xaxis=(None, "x2"),
            layout={"grid": {"rows": 1, "columns": 2, "subplots": [["xy", "x2y"]], "roworder": "bottom to top"}},
            on_range_change="range_change",
            width="100%",
            height="100%",
            selected="{selected_indices}",
        )
```
- a page can be defined without context manager by the add method which could be chained as it return itself
```
page = tgb.Page()
page.add(tgb.text("Hello")).add(tgb.text("World"))
```

dinhlongviolin1 (approved):
LGTM

docs/manuals/gui/page_builder.md
FabienLelaquais: This I will indicate in the documentation for `html()` (which is mainly empty at this point).

docs/manuals/gui/page_builder.md
FabienLelaquais: Rephrased.

docs/manuals/gui/page_builder.md
FabienLelaquais: I addressed the second point.
The first one should in my opinion be an addendum to the chart control, where it makes a lot of sense.

dinhlongviolin1: I agree! As long as the user know that it is an option we are good. We can also add it for selector as they can quickly defines lov.

docs/manuals/gui/page_builder.md
dinhlongviolin1: thanks

PR #683, opened 2023-10-23
Home page

jrobinAV:
image


ghost (approved):
👌

AlexandreSajus (changes requested):
The new text is excellent; it describes perfectly Taipy's current positioning. I would only replace the GIF for scenario management with the one attached. It needs to be more dynamic and showcase more capabilities.

taipy_back

jrobinAV:
> The new text is excellent; it describes perfectly Taipy's current positioning. I would only replace the GIF for scenario management with the one attached. It needs to be more dynamic and showcase more capabilities.

The new screenshot is much better, indeed.

FlorianJacta (approved):

AlexandreSajus (approved):

PR #684, opened 2023-10-23
feat(knowledge-base): implement list filtering script

ghost:
Implement filtering script  for the Knowledge base lists, matching @jrobinAV data-keywords proposal.
It will just require adding a class "tp-filtered" on every `<ul>` that should be filtered by the checkboxes.


jrobinAV:
I would like to settle on a list of filters before merging this PR.

jrobinAV (changes requested):
need to reduce nb of filters

Issue #694, opened 2023-10-29
BUG-Initial public RefMan for 3.0 mixes things up


FabienLelaquais:
**Bug description**

The RefMan as published with Taipy 3.0 is a bit messy, with entities appearing where they don't  belong.

**Expected change**

Clean this up

**Screenshots**

![taipy-avant](https://github.com/Avaiga/taipy-doc/assets/86590727/678291f7-54fc-4254-8ce4-a96218438d5e)

PR #695, opened 2023-10-29
Fix messy RefMan generation

FabienLelaquais:
Addresses #694

### DIFF
--- a/tools/_setup_generation/step_refman.py
+++ b/tools/_setup_generation/step_refman.py
@@ -28,94 +28,6 @@ class RefManStep(SetupStep):
         "taipy.enterprise",
     ]
 
-    # Force API items to be exposed in a given package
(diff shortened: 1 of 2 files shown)


jrobinAV (approved):
The result looks good to me. Much better.

However I am concern about inconsistencies.

In order to avoid asking the user to import too many classes (ScenarioManager, DataManager, JobManager, Orchestrator, etc...), we expose methods as static functions in taipy package. We document in the user manual to use the taipy functions: `taipy.create_scenario(config)`, `taipy.submit(scenario)`, etc.
This is inconsistent regarding the reference manual. 

Nothing can be done now, but I believe we should keep this in mind.

jrobinAV:
It needs to be cherry-picked on top of develop branch as well.

Issue #697, opened 2023-10-31
BUG-Notebook link is dead GUI tutorial


AlexandreSajus:
On this page:
https://docs.taipy.io/en/latest/knowledge_base/tutorials/understanding_gui/#using-notebooks

This link is dead:
![image](https://github.com/Avaiga/taipy-doc/assets/62465003/a4528f61-7e32-41d3-afae-74c2b25e9521)

FabienLelaquais:
This was discussed with @FlorianJacta yesterday.
This problem is due to the migration of the legacy Getting Started into the Knowledge Base structure that forgot about regenerating the notebook document.
I recommend that this generation is performed manually, once, and that the resulting file be stored in the doc repository itself, where it belongs. When the code is updated in the future, so should the notebook.
The generation script, if generic enough, could be stored in the `tools` directory for the doc for future use.

FabienLelaquais:
So I manually created that notebook file, so a minimum of visitors would be impacted.
To me this will be closed at the end of the sprint, unless anyone complains.

Note that this was merged in develop and 3.0 as part of far larger PRs.

Issue #678, opened 2023-10-20
BUG-Wrong cross-refs to Reference Manual


FabienLelaquais:
**Bug description**

Some API elements have the same name although located in different modules.
An example would be `taipy.gui.Page` and `taipy.gui.builder.Page`. When the element name is the only provided information (in a link looking like \`TheElement\^\`), the doc generator postprocessor tries to find the element among all documented ones.
As of today, this search may pick a wrong element.

**Expected change**
There should be a way to specify which package the element should be searched for.

We need to extend the cross-def syntax to allow for specifying as little packages as possible.
In our example, that would be \`(builder.)Page\` to specify which we want.

jrobinAV:
As a workaround, can we use the [`taipy.gui.Page`](../manuals/ref/taipy/gui/page.md) syntax?

FabienLelaquais:
Yes but then it's really not nice in the text flow.

Issue #434, opened 2023-04-20
[Build] Improve warnings reporting


FabienLelaquais:
**Description**

Links to class constructors (__init__()) in the Reference Manual are treated in the post-process phase, introducing warnings messages (Documentation file '*.md' contains a link to '.../[class].__init__()^' which is not found in the documentation files) that get later solved with a message (FIXED cross-ref to '.../[class].__init__()^' from *.md...).

These message make locating actual problems hard.

**Expected change**

Only relevant warnings and errors should be provided.

PR #700, opened 2023-10-31
XRefs to RefMan improvements

FabienLelaquais:
Addressed #434, #678, #694 and #697.

- Allow for duplicate API elements
  (in different packages)
- Rewrite dubious links to constructors
   (recommand the "\`Class,\_\_init\_\_^\`(\`Class\` constructor)" syntax) 
- Fixed some broken links


jrobinAV (approved):
I love it. Non need to think about adding every single method exposed at taipy level as an exception in the refman step !!

Awesome!

jrobinAV (approved):
Some good cleaning!

PR #727, opened 2023-11-09
:memo: Rewrites the intro of the cycles tutorial for clarity purposes

### LINKED ISSUE #378: BUG-The cycle tutorial does not explain to the reader why he should use cycles, and lacks visuals
https://docs.taipy.io/en/develop/knowledge_base/tutorials/cycles_scopes/

The cycles tutorial introduces the purpose of cycles with this. dense block of text:
image

This introduction is too abstract, too broad, and lacks visuals.

Maybe we should introduce the topic by just saying that cycles allow you to:
- share variables between scenarios in the same time frame, this way if I have three prediction scenarios for the month of June, I do not need to create three sales datasets for June but only use the one
- better organize variables in the application (here last_two_years is a shared dataset between both scenarios so it does not appear twice)
image

In general, any feature should have a visual associated to it. For example:
image
Why would I implement this unless it pushes the primary scenario to the top of the scenario_selector list, or unless it adds a star icon to my scenario.

I can work on this, just tell me if there are features of cycles that I am missing

AlexandreSajus:
Adds visuals and simplifies the explanation of the use-case of Cycles in this [tutorial](https://docs.taipy.io/en/develop/knowledge_base/tutorials/cycles_scopes/)

# Before
image

# After
image

Fixes https://github.com/Avaiga/taipy/issues/378


jrobinAV (approved):
Much better, much simpler.

jrobinAV:
Please don't forget to delete the branch when it is merged.

PR #728, opened 2023-11-09
:memo: Rewrites INSTALLATION.MD in a shorter and clearer form

AlexandreSajus:
Multiple people had trouble with building the docs. The installation instructions are currently very complete but also very dense, which leads to confusion. Also, most editors of the docs do not need to know all the options they have for building the doc. I rewrote the instructions in a simple 3-step process:

# Before
image

# After
image


jrobinAV (approved):
Why not.

PR #750, opened 2023-12-11
3rd Party Components tip - From Sankey to Folium

FlorianJacta:
The goal of this PR is to change a tip discussing the integration of third party componentsnusing HTML and Plotly as en example. Right now, the integration with Plotly Python is easy. So, the example changed to integrate Folium.


AlexandreSajus (approved):
Nice! This example is way sexier, we should use this tip by default when creating maps, it is so much simpler

FlorianJacta:
> Nice! This example is way sexier, we should use this tip by default when creating maps, it is so much simpler

It is not using the `content` property now available in Taipy. `content` with Plotly makes it even easier.

PR #765, opened 2023-12-26
Fix file_selector label

jrobinAV:


FabienLelaquais (approved):
Why would you want to change that?
It was not wrong before...
Doesn't hurt, but what's the point?

PR #770, opened 2024-01-04
Add databricks tip

FlorianJacta:
We should add an illustration to the tip @marisogo . When done, it could be merged.


docs/knowledge_base/tips/databricks/example.py
```diff
@@ -0,0 +1,139 @@
...
+
+# main.py
+import taipy as tp 
+
+tp.Core().run()
+
+scenario = tp.create_scenario(scenario_cfg)
+
+scenario.submit()
+print(scenario.result.read())
```
jrobinAV: BReak line at the end of the file.

docs/knowledge_base/tips/databricks/index.md
```diff
@@ -0,0 +1,207 @@
+Integration with other services to process jobs and functions is sometimes necessary.
```
jrobinAV:
```suggestion
Integration with other platforms to execute jobs and functions is sometimes necessary. 
```

docs/knowledge_base/tips/databricks/index.md
```diff
@@ -0,0 +1,207 @@
+Integration with other services to process jobs and functions is sometimes necessary. 
+This tip shows how to use Taipy scenarios with platforms like Databricks simultaneously.
```
jrobinAV:
```suggestion
This tip shows how to integrate Taipy scenarios with platforms like Databricks. 
```

docs/knowledge_base/tips/databricks/index.md
```diff
@@ -0,0 +1,207 @@
...
+        # Initialization details...
+
+    @property
+    def cluster_id(self) -> str:
+        # Retrieve cluster ID...
+
+    def status(self):
+        # Check cluster status...
+
+    def get(self, endpoint, dataset=None, timeout=BASIC_TIMEOUT):
```
jrobinAV: I believe `get` is a wrong name or at least misleading. Here, the meaning of get is that you want to get the endpoint, which is not at all what you are doing.

You can use run, run_job, compute, process, trigger_execution, execute, compute_ressults, compute_and_get_ressults...

docs/knowledge_base/tips/databricks/index.md
```diff
@@ -0,0 +1,207 @@
...
+            return None
+
+```
+
+As you can see, multiple values are used to connect to Databricks and to the right job.
+
+- *JobEndpoint*: the name of the job.
+- *DatabricksBearerToken*: your bearer token. Create one using this [tutorial](https://docs.databricks.com/en/dev-tools/auth/pat.html).
+- *DatabricksEndpoint*: Databricks endpoint which is the base URL of your Databricks (example: *xxxyyyyzzz.azuredatabricks.net*).
+- *DatabricksClusterName*: the name of the cluster to be used.
```
jrobinAV: Why do you use environment variables? 
Since it is just an example/tip and not real code, I would just use simple variables for better readability.
Eventually, you can leave a comment explaining the values that could be retrieved from environment variables, at least for the token.

docs/knowledge_base/tips/databricks/index.md
```diff
@@ -0,0 +1,207 @@
...
+scenario_cfg = Config.configure_scenario("scenario", task_config=[task_databricks_cfg])
+```
+
+Now, that the scenario is configured, it can be created and executed to retrieve the 
+proper results.
+
+```python
+import taipy as tp 
+
+tp.Core().run()
```
jrobinAV: Please use the `if __name__ == "__main__"` here.

docs/knowledge_base/tips/databricks/index.md
FlorianJacta: I want people to avoid trying the code and seeing that it is not working because the secrets do not work, so I want to make it obvious to use their variables. Also, this is a generic code that could work for anyone. What do you think?

docs/knowledge_base/tips/databricks/index.md
FlorianJacta: There are two steps in the `get` function. One is where you trigger the job, and one is where you get the endpoint many times at some interval to get the results. I can go with `run_and_get_results` if you like it

docs/knowledge_base/tips/databricks/index.md
jrobinAV: For me, it is strange to pass the databricks job, the databricks cluster name and its endpoint as env variables, especially for the job. Plus it slightly degrades the readability.
However, you have very good arguments as well. So I let you decide.

jrobinAV (approved):

PR #771, opened 2024-01-04
Removes src import in page_builder page

AlexandreSajus:
Removes src import on this page:
image

https://docs.taipy.io/en/latest/manuals/gui/page_builder/


FabienLelaquais (approved):
👌

Issue #693, opened 2023-10-28
BUG-Document installation on Colab


FabienLelaquais:
**Bug description**

[This issue](https://github.com/Avaiga/taipy/issues/365) mentions that Taipy 3.0 cannot be installed on Colab, and the Taipy doc doesn't help.

**Expected change**

The documentation must explicitly state the workaround on the installation page, maybe in the related Tip section.

PR #797, opened 2024-01-22
Add Colab and Python version

FlorianJacta:
Maybe related to: https://github.com/Avaiga/taipy/issues/730

Resolves: https://github.com/Avaiga/taipy/issues/731

Resolves: https://github.com/Avaiga/taipy/issues/734

Resolves: https://github.com/Avaiga/taipy-doc/issues/693

image

PR #798, opened 2024-01-22
Merge pull request #797 from Avaiga/doc/fix-installation-guide

FlorianJacta:
Maybe related to: https://github.com/Avaiga/taipy/issues/730

Resolves: https://github.com/Avaiga/taipy/issues/731

Resolves: https://github.com/Avaiga/taipy/issues/734

Resolves: https://github.com/Avaiga/taipy-doc/issues/693

image

### DIFF
--- a/docs/installation/index.md
+++ b/docs/installation/index.md
@@ -10,7 +10,7 @@ powerful web applications.
 
 # Prerequisite
 
-Before installing Taipy, ensure you have Python (version 3.8 or above) and 
+Before installing Taipy, ensure you have Python (**between version 3.8 and 3.11**) and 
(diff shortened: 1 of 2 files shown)


jrobinAV (approved):

docs/installation/index.md
jrobinAV, line 1: In terms of priority, I would move the colab installation after the conda installation.

FlorianJacta: True, but we never had anyone complain about Conda, as they normally know how to use it. However, we had complaints about Colab. What do you think?

FlorianJacta: A normal installation of Taipy will not work on Colab (`!pip install --ignore-installed taipy`) and a normal use of Taipy (just running the application) will also not work

PR #832, opened 2024-02-13
Document the format property of the date and date_range controls.

### LINKED ISSUE #772: Add examples for date format in docs
Both date and date_range visual elements' docs do not provide examples for the format property which is supposed to define the date format. This leads to users not being able to choose the date format.
https://docs.taipy.io/en/develop/manuals/gui/viselements/date/
https://docs.taipy.io/en/develop/manuals/gui/viselements/date_range/

FabienLelaquais:
Resolves [taipy#772](https://github.com/Avaiga/taipy/issues/772).


docs/manuals/gui/viselements/date.md_template
```diff
@@ -15,18 +25,86 @@ The [Stylekit](../styling/stylekit.md) provides a specific class that you can us
...
+```python
+import datetime
+
+date = datetime.datetime(1789, 7, 14, 17, 5, 12)
+```
+
+You can create a date selector that represents it with the following definition:
+
+!!! taipy-element
+    default={date}
```
FredLL-Avaiga, line 48: default ? or date ?

FredLL-Avaiga: or is it some magic to generate the right example in different formats ?

FredLL-Avaiga (approved):
never mind my comment

docs/manuals/gui/viselements/date.md_template
FabienLelaquais: Indeed there is.

PR #865, opened 2024-03-07
Add documentation for part.content and content providers

### LINKED ISSUE #739: Document the content property of the part block, and content providers.
**Description**

The next version of Taipy GUI provide a new property to the `part` control that makes it possible to integrate, in an `iframe`, HTML code generated by a function (that we call a *content generator*).
Content generators return an value of any kind (such as a string, or an object) that can be converted to HTML using registered *content providers*: users can define content providers for any type (returned by the content generator).

Predefined implementations are available for Plotly Express and Matplotlib.

**Expected change**
[] The doc reflects the new feature.

FabienLelaquais:
Resolves #739.


FredLL-Avaiga (approved):
Looked good

PR #873, opened 2024-03-14
More details on HTML support in Taipy GUI

### LINKED ISSUE #776: Html pages drop the event handlers on HTML elements
**Description**

XSS issues prevent the page generation to keep event handlers on pure HTML elements.

Another thing that needs documentation is the fact that HTML pages must rely on the XHTML syntax.

**Expected change**

Document the limitation and the workaround:
```
from taipy.gui import Gui, Html

page = Html(
    """
<!DOCTYPE html>
<html>
<head>
    <title>Simple Webpage with JavaScript</title>
    <script type="text/javascript">
        function ze_function() {alert("Hello, this is a simple JavaScript alert!")};
        const ze_button = document.getElementById('ze_button');
        ze_button.addEventListener('click', ze_function);
    </script>
</head>
<body>
    <h1>Welcome to My Webpage</h1>
    <button id="ze_button">Click Me!</button>
</body>
</html>
"""
)

Gui(page).run()
```

FabienLelaquais:
Resolves #776.

### DIFF
--- a/docs/manuals/gui/pages/html.md
+++ b/docs/manuals/gui/pages/html.md
@@ -52,7 +52,8 @@ Here are the rules that the page definition must follow:
 
 # Event handlers
 
-Taipy GUI drops the event handlers explicitly set using native HTML elements' attributes.
+You cannot declare event handlers explicitly in HTML elements because Taipy GUI cannot use the event
+handlers' attribute value.
 
 Here is an example of HTML content in this situation. A Taipy application creates a page entirely
 defined in HTML as follows:
--- a/docs/manuals/gui/pages/html.md
+++ b/docs/manuals/gui/pages/html.md
@@ -6,8 +6,8 @@ to describe pages, so you have all the power of the HTML grammar to organize you
 
 You do not need to create the header and body parts, as Taipy takes care of this for you.
 
-Creating a page that displays HTML content is simple:
-
+Creating a page that displays HTML content needs the application to create an instance of the
+`Html^` class:
 ```python
 from taipy.gui import Html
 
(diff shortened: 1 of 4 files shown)


docs/manuals/gui/pages/html.md
```diff
@@ -16,12 +16,103 @@ html_page = Html("""
...
+""")
+
+Gui(page).run()
+```
+
+In line 16, you can spot where the HTML button is created, along with its event handler: the
+function *send_the_alert()*, defined in line 9, is called when the user presses the button.
+
+But if you run this application and press the button, you will *not* witness the invocation of the
+function. That is because when Taipy converts the HTML input into the page sent to the browser, it
```
FredLL-Avaiga, line 88: this is not true
if you look at the browser's console you'll see that
- onclick => onClick
- onClick="send_alert();" => jsxParser complains that that property should be a function and not a string

if you write it like this onClick="{!send_alert!}", the parser does not complain (that's the way we use to pass js objects to jsxParser.

but ... it's not working with nothing in the console

docs/manuals/gui/pages/html.md
```diff
@@ -16,12 +16,103 @@ html_page = Html("""
...
+removes the event handlers of the HTML elements.
+
+To fix this problem, you must set the event handler by code when the page is executed on the client.
```
FredLL-Avaiga: thefix works but the reason is not right

FredLL-Avaiga (changes requested):
the reason for js handler ios not right but I can't figure out why :-(

docs/manuals/gui/pages/html.md
FabienLelaquais: I'm trying to not expose users to JSX.
Note how the term "React" appears basically nowhere (except in the Extension API of course).

docs/manuals/gui/pages/html.md
FabienLelaquais: I'll find a less precise way to mention this...
Thx

docs/manuals/gui/pages/html.md
```diff
@@ -16,12 +16,103 @@ html_page = Html("""
...
+- Strict adherence to syntax rules is required. All elements must be properly nested, closed, and in
+  lowercase.
+    - All the tag names must be lowercase.
+    - All the attribute names must be lowercase, and all values must be quoted.<br/>
+      Note that the XHTML syntax parsed by Taipy GUI in the context of a visual element allows for
+      leaving an attribute valueless. The value for the property is then set to True.
+
+# Event handlers
+
+Taipy GUI drops the event handlers explicitly set using native HTML elements' attributes.
```
dinhlongviolin1: This sentence doesnt sit right with me. Could not wrap my head around what it is trying to convey.

dinhlongviolin1 (approved):
LGTM

FredLL-Avaiga (approved):

docs/manuals/gui/pages/html.md
FabienLelaquais: Better in the last commit?

dinhlongviolin1 (approved):
LGTM

PR #931, opened 2024-04-24
Doc/taipy#410 - Update export_scenario() and import_scenario() user manual

### LINKED ISSUE #410: Possbility to import a scenario (after an export)
**What would that feature address**
The possibility of importing a scenario/scenarios after exporting it with Taipy. The problem is that in the context of a project, different scenarios are present in different environments (Test, QA, Production). 

For example, the end user tests its application on the Test environment and creates meaningful scenarios. When the test is done, the application goes into QA and Production. The end user wants the scenarios he created in the other environments.

Another use case could be if the end-user encounters an issue with a scenario/scenarios he has created; he would then want to send it/them to its technical team (Data Scientist, Python Developers, ...) to debug them.

***Description of the ideal solution***
This export/import system should work with the different storage methods (Filesystem, SQL, Mongo). It could be another option of the CLI.
The goal would be to have the possibility to import a scenario. The difficulty comes from ensuring the consistency of the import regarding the entities related to the scenario and the data attached to data nodes.

**Acceptance Criteria**
- [ ] Ensure new code is unit tested, and check code coverage is at least 90%
- [ ] Create related issue in taipy-doc for documentation and Release Notes
- [ ] Check if a new demo could be provided based on this, or if legacy demos could be benefit from it
- [ ] Ensure any change is well documented

trgiangdo:
Resolves https://github.com/Avaiga/taipy/issues/410

### DIFF
--- a/docs/manuals/core/entities/scenario-cycle-mgt.md
+++ b/docs/manuals/core/entities/scenario-cycle-mgt.md
@@ -231,13 +231,20 @@ be raised.
 # Export a scenario
 
 You can export a scenario with its related entities in `JSON` format into a folder using `taipy.export_scenario()^`.
-This method takes as parameters the scenario id and the path to the export folder.
+This method takes as the parameters:
 
-!!! warning
+- *scenario_id* represents the id of the scenario to export.
+- *folder_path* represents the path to the export folder.
+  If the path exists and the override parameter is False, the `ExportFolderAlreadyExist^` exception is raised.
+- *override* (bool): is a boolean indicating if the export folder will be overridden when it exists. The
+  default value is False.
+- *include_data* is a boolean indicating if the data files of file-based data nodes are exported. File-based
+  data nodes include Pickle, CSV, Excel, Parquet, and JSON data nodes. If the scenario has a data node that
+  is not file-based, a warning will be logged, and the data of that data node will not be exported. The
+  default value is False.
 
-    The `folder_path` will be **overwritten** if it does exist.
-
-Alternatively, you can use the `Scenario.export()^` method.
+Alternatively, you can use the `Scenario.export()^` method and provide the same parameters except for the
+*scenario_id* parameter.
 
 ```python linenums="1"
 import taipy as tp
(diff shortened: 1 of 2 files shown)


docs/manuals/core/entities/scenario-cycle-mgt.md
```diff
@@ -231,23 +231,56 @@ be raised.
...
+scenario.export(folder_path="./monthly_scenario", override=True, include_data=True)
+```
+
+!!! note
+
+    The entity in the export folder may not work correctly due to the data path of the exported
+    file-based data nodes.
+
+    For the scenario to work correctly, please import the scenario into the current Taipy application
+    using the `taipy.import_scenario()^` method.
```
jrobinAV: I don't understand why we need this note. importing the scenario should be the only way to make it work. no? What does it mean "The entity in the export folder may not work correctly" ?

docs/manuals/core/entities/scenario-cycle-mgt.md
```diff
@@ -231,23 +231,56 @@ be raised.
...
+
+# Import a scenario
+
+You can import a scenario with its related entities and data files from a folder in `JSON`
+format using `taipy.import_scenario()^` into the current version of the Taipy application.
+This method takes as the parameters:
```
jrobinAV:
```suggestion
You can import a scenario that has been exported before using the `taipy.import_scenario()^` method. This method takes as the parameters:
```

docs/manuals/core/entities/scenario-cycle-mgt.md
trgiangdo: We never said that importing is the only way to make it works.

Besides, the output of the `export()` method is a folder that is really similar to a `.taipy` folder, so there may be a misunderstanding

jrobinAV: Oh, you believe some people could just put the files resulting from the export in the `.taipy` folder and try to run the app?
I get it now.

jrobinAV (approved):
Neat!

PR #939, opened 2024-05-03
Relnotes/taipy#1198 - Fix DataNode.is_up_to_date error

### LINKED ISSUE #1198: BUG- <DataNode>.is_up_to_date raises error when never written before
**Description**

The `<DataNode>.is_up_to_date` method breaks when the Data Node has never been written.

```
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "c:\Users\jacta\.conda\envs\bar-cutting\lib\site-packages\taipy\core\data\data_node.py", line 528, in is_up_to_date
    and ancestor_node.last_edit_date > self.last_edit_date
TypeError: '>' not supported between instances of 'datetime.datetime' and 'NoneType'
```

**How to reproduce**

```python
from taipy.config import Config
import taipy as tp

Config.configure_job_executions(mode="standalone")


# Normal function used by Taipy
def double(nb):
    return nb * 2


# Configuration of Data Nodes
input_cfg = Config.configure_data_node(id="somedata", default_data=21)
output_cfg = Config.configure_data_node(id="result")


# Configuration of tasks
first_task_cfg = Config.configure_task(id="double",
                                       function=double,
                                       input=input_cfg,
                                       output=output_cfg)


# Configuration of scenario
scenario_cfg = Config.configure_scenario(id="my_scenario",
                                         task_configs=[first_task_cfg],
                                         name="my_scenario")


if __name__=="__main__":
    tp.Core().run()
    scenario_1 = tp.create_scenario(scenario_cfg)

    print(scenario_1.result.is_up_to_date)
```

**Expected behavior**
We should not get an error; the result should be False if the Data Node is invalid.

Possible solution:

Change the *is_up_to_date* function in *taipy/core/data/data_node.py*, line 510:

```python
@property
def is_up_to_date(self) -> bool:
    """Indicate if this data node is up-to-date.

    Returns:
        False if a preceding data node has been updated before the selected data node
        or the selected data is invalid.<br/>
        True otherwise.
    """

    if self.is_valid:
        from ..scenario.scenario import Scenario
        from ..taipy import get_parents

        parent_scenarios: Set[Scenario] = get_parents(self)["scenario"]  # type: ignore
        for parent_scenario in parent_scenarios:
            for ancestor_node in nx.ancestors(parent_scenario._build_dag(), self):
                if (
                    isinstance(ancestor_node, DataNode)
                    and ancestor_node.last_edit_date
                    and ancestor_node.last_edit_date > self.last_edit_date
                ):
                    return False
    return False

```

**Runtime environment**
 - Taipy: 3.1.1

**Acceptance Criteria**
- [ ] Ensure new code is unit tested, and check code coverage is at least 90%
- [ ] #1200

trgiangdo:
Resolves https://github.com/Avaiga/taipy/issues/1198.

This PR notices the fix on the relnote.

### DIFF
--- a/docs/relnotes.md
+++ b/docs/relnotes.md
@@ -66,7 +66,10 @@ This is the list of changes to Taipy releases as they were published.
 
 <h4><strong><code>taipy-gui</code></strong> 3.2.0</h4>
 
-<h4><strong><code>taipy-core</code></strong> 3.2.1</h4>
+<h4><strong><code>taipy-core</code></strong> 3.2.0</h4>
+
+- `DataNode.is_up_to_date^` raises an error when the data node has never been written.<br/>
+  See [issue #1198](https://github.com/Avaiga/taipy/issues/1198).
 
 # Enterprise edition: 3.2
 


jrobinAV (approved):
Neat !

PR #947, opened 2024-05-07
Fix/#895 - The ellipsis should link to manuals/usermans/

### LINKED ISSUE #895: BUG-404 page when clicking on ellipsis
**Bug description**
Clicking on the ellipsis on the end of the Home page of Taipy leads to 404.

**Expected change**
This should redirect to the correct page.

**Screenshots**
image

trgiangdo:
Resolves #895

The ellipsis now would link to https://docs.taipy.io/en/latest/manuals/usermans/

We can also consider linking to https://docs.taipy.io/en/latest/manuals/, which shows both User Manual and Reference Manual. Let me know what you think @FabienLelaquais @jrobinAV

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -141,7 +141,7 @@ from initial prototypes to production-ready applications.
     </a>
   </li>
     <li>
-    <a class="tp-pill" href="manuals/about/">
+    <a class="tp-pill" href="manuals/usermans/">
       <span>…</span>
       <div class="tp-tooltip">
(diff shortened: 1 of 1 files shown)


jrobinAV (approved):
The link is about taipy functionalities. I believe the user manual is indeed a better target than reference manual.

Once this PR is merged, can you cherry-pick the commit on 
`release/3.1`, and `release/3.0` branches?

FabienLelaquais (approved):
👌

PR #967, opened 2024-05-19
Fix most generation warnings and errors related to links

FabienLelaquais:
Reviews are people related to modified files.


FredLL-Avaiga (approved):

jrobinAV (approved):
Neat! 

I need to propagate these to my PR on the userman.
@FabienLelaquais Do you plan to have more changes to the userman? If yes, can we do it in this PR as well?

FabienLelaquais:
> Neat!
> 
> I need to propagate these to my PR on the userman. @FabienLelaquais Do you plan to have more changes to the userman? If yes, can we do it in this PR as well?

I have a bunch, but I won't commit before next sprint, as we discussed last week,
Thanks

PR #985, opened 2024-06-11
Add support for hiding a property's doc

### LINKED ISSUE #977: Support the hiding of element properties using viselements.json
**Description**
That would allow to hide a property from any element's documentation.

**Proposed solution**

The best place to indicate that a given property, for a given element, should be hidden, is the `viselements.json` file where it is declared.
This file is used for generating the documentation, for the auto-completion feature of Visual Studio GUI, and the generation of the Page Builder API.

*Non inherited property*
In order to hide an element's property, we propose to simply prefix the documentation content with the string "UNDOCUMENTED".

*Inherited property*
Inherited properties (such as `id` or `hover_text`) cannot use this mechanism because the element property's doc attributes are merged.

We propose to add an attribute called `hide` in the JSON description.
`hide` would be of type `boolean` and default to `false`.

**Expected change**
```
                    {
                        "name": "hover_text",
                        "type": "dynamic(str)",
                        "doc": "The information that is displayed when the user hovers over this element."
                    }
```
would be declared in `indicator` (who inherit from `shared`) as 
```
                    {
                        "name": "hover_text",
                        "hide": true
                    }

```

FabienLelaquais:
Resolves #977.

Any property of any element can have its documentation dropped.
A similar work has to be done in the [TPB API generation](https://github.com/Avaiga/taipy/issues/1395) and in the [Visual Studio Taipy GUI extension](https://github.com/Avaiga/taipy-studio-gui/issues/54).

### DIFF
--- a/tools/_setup_generation/step_viselements.py
+++ b/tools/_setup_generation/step_viselements.py
@@ -89,6 +89,7 @@ class VisElementsStep(SetupStep):
                     new_elements[element_type] = element_desc
             self.elements.update(new_elements)
             # Find default property for all element types
+            # and remove hidden properties.
             for element_type, element_desc in new_elements.items():
                 default_property = None
                 if properties := element_desc.get(__class__.PROPERTIES, None):
@@ -97,6 +98,8 @@ class VisElementsStep(SetupStep):
                             if property[__class__.DEFAULT_PROPERTY]:
                                 default_property = property[__class__.NAME]
                             del property[__class__.DEFAULT_PROPERTY]
+                        if property.get("hide", False):
+                            property["doc"] = "UNDOCUMENTED"
                 element_desc[__class__.DEFAULT_PROPERTY] = default_property
 
             # Resolve inheritance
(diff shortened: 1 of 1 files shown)


tools/_setup_generation/step_viselements.py
```diff
@@ -246,6 +249,9 @@ def generate_element_doc(element_type: str, element_desc: Dict, prefix: str):
             ):
                 property_desc = property_descs[property_name]
                 name = property_desc[__class__.NAME]
+                doc = property_desc.get("doc", None)
+                if doc.startswith("UNDOCUMENTED"):
```
FredLL-Avaiga, line 253: that will break if doc is None

FredLL-Avaiga (approved):

tools/_setup_generation/step_viselements.py
FabienLelaquais: So very True.

FredLL-Avaiga (approved):

PR #987, opened 2024-06-13
Adds new LLM demos and multi-user tuto to LATEST

AlexandreSajus:
Pushes the new LLM demos and the multi-user tutorial to latest so it is visible in the docs and we can do marketing on it

This is the first time I am doing a push to LATEST so please check that I did not do any shenanigans that would break the docs. thank you very much 👍👍👍


FlorianJacta (approved):
👌

PR #1050, opened 2024-07-24
Feature/#986 - Document best practice on main script

### LINKED ISSUE #986: Protect main script from being executed outside the main process
Taipy can start subprocesses using a 'spawn' method. Because of that, the main module is re-executed every time a sub-processed is spawned. In order NOT to re-execute some code that should not be re-executed, the good practice Taipy wants to enforce is to add the boilerplate
condition `if __name__ == "__main"` at the top of the main script.

- [ ] Ensure the `if __name__ == "__main"` condition is explicitly documented and explained in the standalone execution mode documentation (user, ref, tuto, etc.) The motivation should be explicit.
- [ ] Ensure every complete example includes the `if __name__ == "__main"` in all chapters

trgiangdo:
Resolves https://github.com/Avaiga/taipy-doc/issues/986

The warning for standalone mode is added in:
- `docs/manuals/userman/configuration/job-config.md`
- `docs/tutorials/scenario_management/5_job_execution/index.md`


docs/manuals/userman/configuration/job-config.md
```diff
@@ -85,10 +85,17 @@ You can configure the *standalone* mode with the following config:
 To execute multiple `Job^`s simultaneously, you can set the *max_nb_of_workers* to an
 integer value greater than 1. That starts each `Job^` in a dedicated process with
 *max_nb_of_workers* as the limit of concurrent processes that can run simultaneously.
+The default value of *max_nb_of_workers* is 2.
 
-!!! note
+!!! warning
 
-    The default value of *max_nb_of_workers* is 2.
+    In *standalone* mode, when a subprocess is spawned, the code is re-executed except
```
FlorianJacta, line 92: What about the code inside other pages?

trgiangdo: It will be rerun, but if the Core and GUI service are protected in the `if __name__ == "__main__"`, it won't matter since the subprocess will only execute the tasks

FlorianJacta (approved):

PR #1053, opened 2024-07-25
Propagates decimator tutorial tgb approach to latest

AlexandreSajus:
image

### DIFF
--- a/docs/tutorials/large_data/3_decimator/index.md
+++ b/docs/tutorials/large_data/3_decimator/index.md
@@ -69,21 +69,40 @@ We evaluated three different algorithms:
 
 Here is an example of code showing how to use the Min-Max Decimator:
 
-```python
-import yfinance as yf
-from taipy.gui import Gui
-from taipy.gui.data.decimator import MinMaxDecimator, RDP, LTTB
+=== "Markdown"
+    ```python
+    import yfinance as yf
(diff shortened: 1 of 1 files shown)


FlorianJacta (approved):
LGTM

FabienLelaquais:
@AlexandreSajus I really don't like this syntax.
We need to allow (and use) `decimator=decimator_instance` and *not* force the use of a variable.
I don't know if this works right away but I suspect it does.

AlexandreSajus:
> @AlexandreSajus I really don't like this syntax. We need to allow (and use) `decimator=decimator_instance` and _not_ force the use of a variable. I don't know if this works right away but I suspect it does.

Unfortunately, the syntax you propose currently does not work both in Taipy latest and in Taipy develop. I just tested both.

FabienLelaquais:
> > @AlexandreSajus I really don't like this syntax. We need to allow (and use) `decimator=decimator_instance` and _not_ force the use of a variable. I don't know if this works right away but I suspect it does.
> 
> Unfortunately, the syntax you propose currently does not work both in Taipy latest and in Taipy develop. I just tested both.

Thank you @AlexandreSajus. We MUST fix that. I have created [this issue](https://github.com/Avaiga/taipy/issues/1591).

PR #1064, opened 2024-07-31
Restructure doc for new build of tutos and gallery

### LINKED ISSUE #1063: Structure and UI are interconnected inside Gallery and Tutorials
**Description**
Currently, the UI of the tutorials and gallery is determined by the structure of these sections. An application in the gallery must reside in the corresponding section folder (e.g., "finance") to appear in that section. Consequently, if the label (such as "finance") changes, all dependent URLs must also change, leading to frequent URL breaks. Additionally, reorganizing these sections requires modifying both the gallery structure and the URLs.

The order in which applications appear in a section's (e.g., finance) index file is linked to its URL, making the system inflexible. The sections serve more as filters for applications rather than providing a proper description, so being in one section or another should not have a huge impact like this.

**Expected Change**
The structure and UI should be separated to enhance flexibility and prevent disruptions.

FlorianJacta:
Resolves: #1063 

Let's discuss of having this kind of solution. Please let me know what your thoughts are.

Nothing has changed for the user except that now, users will see this in the URL of demos and tutorials: docs.taipy.io/en/develop/gallery/articles/chatbot/.

Now the order is set by the *order* property of the *index.md* of the related demo or tutorial.

For the developer, now everything is in the same place, which is worse in visualizing where the articles are. However, it is simpler to contribute to the documentation, reorder the documentation and change sections of the doc.


AlexandreSajus:
This is a good idea since ordering demos before required adding a number to the folder name. I see now that you use an order property in the index header of each demo. How is the ordering implemented? If this works, I'm okay with it. I'll test it once broken links issues are addressed and let R&D judge the implementation.

jrobinAV (approved):

PR #1072, opened 2024-08-05
Doc/Enterprise#431 - Move production version to enterprise

trgiangdo:
Resolves https://github.com/Avaiga/taipy-enterprise/issues/431

### DIFF
--- a/docs/manuals/cli/manage-versions.md
+++ b/docs/manuals/cli/manage-versions.md
@@ -61,7 +61,9 @@ $ taipy manage-versions --list
 Version number                         Mode                   Creation date
 d74ec95e-6b98-4612-b50b-d171599fa3e9   Development (latest)   2023-01-19 14:45:10
 3.0                                    Experiment             2023-01-18 12:10:55
+2.0                                    Production             2023-01-16 15:10:41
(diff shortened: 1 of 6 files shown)


docs/manuals/userman/versioning/index.md
```diff
@@ -34,6 +34,10 @@ a Taipy application.
   algorithms, investigating the impacts of a parameter change, etc. For more information, please refer to
   [experiment mode](experiment_mode.md).
 
+!!! warning "Available in Taipy Enterprise edition"
+
+    The production mode is relevant only to the Enterprise edition of Taipy.
+
```
jrobinAV: I would move this within the third bullet point, probably after it.

docs/manuals/cli/manage-versions.md
jrobinAV, line 1: I would keep the example with the production versions. You already added warnings saying that production versions are only available in taipy enterprise edition.

docs/manuals/userman/versioning/index.md
trgiangdo: it is now moved inside the bullet point

jrobinAV (approved):

PR #1098, opened 2024-08-28
Replace Core service by Orchestrator Service

jrobinAV:
adresses https://github.com/Avaiga/taipy/issues/1567

We missed a lot of Core terms in the documentation. 
I hope these are the last ones.


FlorianJacta (approved):
Seems great!

PR #1100, opened 2024-08-28
Makes Databricks docs public

AlexandreSajus:
Makes Databricks documentation public:
image

This documentation uses Taipy Enterprise objects. This is disclosed within the article. We have to make this documentation public as it is a requirement to become a Databricks partner. This should be fine since none of the article's contents can be used using Taipy Community.


FlorianJacta (approved):
LGTM

AlexandreSajus:
Creating a new pull request since the repo structure has changed and it is causing merge conflicts

PR #1101, opened 2024-08-28
correcting single_client parameter value from True -> False in the ca…

specialkapa:
Simply made the change suggested in [issue:#1702](https://github.com/Avaiga/taipy-doc/issues/1139)


FlorianJacta (approved):
It seems good to me, thanks!

jrobinAV:
Thank you very much!

specialkapa:
nae bother! it is my pleasure :D

PR #1103, opened 2024-08-29
Makes Databricks docs public

AlexandreSajus:
Same as https://github.com/Avaiga/taipy-doc/pull/1100 but fixes merge conflicts

Makes Databricks documentation public:
image

This documentation uses Taipy Enterprise objects. This is disclosed within the article. We have to make this documentation public as it is a requirement to become a Databricks partner. This should be fine since none of the article's contents can be used using Taipy Community.


docs/tutorials/articles/databricks/index.md
```diff
@@ -3,29 +3,205 @@ title: Integrating Databricks
 category: integration
 data-keywords: scenario task
 short-description: A guide to integrate Databricks with Taipy scenarios.
-order: 24
```
FlorianJacta: You should not delete the order line here

FlorianJacta (changes requested):
Add order to the article

FlorianJacta (approved):

PR #1105, opened 2024-08-30
Doc/#1007 improve navigability

### LINKED ISSUE #1007: Improve doc navigability and UX
**Description**
- Collect feedback from users on discord
- Analyze feedback
- Make a proposition improve navigability as a sketch design, a drawing, or a POC

FlorianJacta:
Resolves: #1007

You can see the POC for this PR here: https://docs.taipy.io/en/doc-add-sections-and-quick-access/ (the links are not correct, and the section on visual elements is different; in this PR (1105), all visual elements are listed without putting them into groups, I propose to do that in a second time)

What is done:
- Quick access in html and CSS for quick access
- '/' in search bar
- Top header erased
- Boost Viz elements
- Automatic generation of Viz elements doc
- CSS for navbar highlight
- Refactor manuals to user manual

Needs to be done:
- Categorized viz elements in Controls, SM controls, and blocks
- Where to put Releases Notes and Contributing


RymMichaut:
Hey Flo,
Nice job.

Can we add a  〉at the beginning of the charts to show that it's a drop-down menu (like in Tutorials)? It feels like the menu level is not the same without it. It would standardize the navigation.
image

- And when dropping down the "charts", can we indent its drop-down menu?
image

- The new charter has a new palette: #ff371a  #0b0b0e  #55c1f6  #ffffff. We may use the same

FlorianJacta:
Thank your @RymMichaut !

> Can we add a 〉at the beginning of the charts to show that it's a drop-down menu (like in Tutorials)? It feels like the menu level is not the same without it. It would standardize the navigation.

True! In the current PR (not the view you are seeing), there is no arrow at all; you cannot expand the chart section. This will come later as the code is a bit hard to automate.

> And when dropping down the "charts", can we indent its drop-down menu?

I'm not sure I understand this, but I think this will also be done with the next PR.


> The new charter has a new palette: #ff371a #0b0b0e #55c1f6 #ffffff. We may use the same

That would be best, but this will come in another PR, too 😄 . This change impacts Taipy Doc and Taipy.

jrobinAV (approved):
Much better.

PR #1112, opened 2024-09-06
Adds Databricks Partner Token to docs

AlexandreSajus:
Adds Databricks Partner Token image to Databricks integration tutorial:
image


FlorianJacta:
Maybe, we could add this svg somewhere as visible as it is now but less big

AlexandreSajus:
Changes the layout so the image is placed better:
image

FlorianJacta (approved):
LGTM

PR #1169, opened 2024-10-17
Add supply chain monitoring demo

arcanaxion:
/gallery:
gallery

/gallery/articles/supply_chain_monitoring:
supply_chain_monitoring


arcanaxion:
@FlorianJacta @AlexandreSajus

docs/gallery/articles/supply_chain_monitoring/index.md
```diff
@@ -0,0 +1,69 @@
+---
+title: Supply Chain Monitoring
+category: decision_support
+data-keywords: dashboard maps enterprise
+short-description: Supply Chain Monitoring dashboard presented as a simulation.
+img: supply_chain_monitoring/images/home_page.png
```
FlorianJacta, line 7: You have to put an `order` to your article and it should correspond where it should be seen in the index pages and it should be coherent on where it is in the navigation too

mkdocs.yml_template
```diff
@@ -258,6 +258,7 @@ nav:
       - "Optimize Bar Cut Sizes": gallery/articles/bar_cutting/index.md
       - "Churn Classification": gallery/articles/churn_classification/index.md
       - "Drift Detection": gallery/articles/drift_detection/index.md
+      - "Supply Chain Monitoring": gallery/articles/supply_chain_monitoring/index.md
```
FlorianJacta: Make sure that this corresponds to the order the article is found in the index pages

FlorianJacta (changes requested):

docs/gallery/articles/supply_chain_monitoring/index.md
arcanaxion: Ah, I used the `supply_chain` article as a reference which didn't have that metadata. I don't really understand what you mean though. Can you give me an example? E.g. `supply_chain_3_echelons` has order=3, but I don't understand how this affects its position

FlorianJacta: The order is the order that the demos are appearing in the index pages with their cards. Supply chain 3 echelon should be the third in the index page: 
image


It seems that he is 4 but because supply-chain is not correct and has no order. (which is a bug bu having it first is correct for supply chain)

arcanaxion: Okay, I think I understand. So since the current final article (tweet_generation) is order=23, I can put this as order=24? Or alternatively, if I want to insert it somewhere, I need to increment the order for every article after it?

Let's say I set this article as order=24, what am I supposed to change in `mkdocs.yml_template`?

FlorianJacta: You should increment the order of every article because I guess we want it to appear in the middle or top of the index when you go to Gallery. If you want to put it at number 10, you should increment all the article order >= 10

mkdocs.yml_template
FlorianJacta: Then, you have to put this line in the order it is represented when you go to the decision support index page

arcanaxion: If I'm understanding correctly, that means that the [current docs](https://docs.taipy.io/en/develop/gallery/decision_support/) is "wrong", right? The left nav is not in the same order that it appears on the decision support index page

image

FlorianJacta: Yep, it is wrong. Check the other sections; they have the correct order. Don't know why it is like this for this one...

arcanaxion: Is it ok for me to just fix it in this PR?

FlorianJacta: Sure, go ahead! Thanks!

arcanaxion:
Article orders are as follows:

```
('supply_chain', 1)
('stock_visualization', 2)
('covid_dashboard', 3)
('supply_chain_3_echelons', 4)
('qsr_sales_forecasting', 5)
('sales_dashboard', 6)
('supply_chain_monitoring', 7)
('production_planning', 8)
('chatbot', 9)
('industry_agglom', 10)
('commodity_trading', 11)
('drift_detection', 12)
('vision_chatbot', 13)
('rag_chatbot', 14)
('background_remover', 15)
('talk_to_taipy', 16)
('bar_cutting', 17)
('churn_classification', 18)
('face_recognition', 19)
('fraud_detection', 20)
('image_classif', 21)
('movie_genre_selector', 22)
('pollution_sensors', 23)
('sentiment_analysis', 24)
('tweet_generation', 25)
```

FlorianJacta (approved):
LGTM

Issue #1170, opened 2024-10-18
Inject installation documentation from the INSTALLATION.md file in taipy repo


jrobinAV:
Create a new step in setup generation to inject the installation documentation.

PR #1175, opened 2024-10-21
Feature/#1170 factorize installation contributing doc

jrobinAV:
resolves issue #1170 

Should be reviewed along with https://github.com/Avaiga/taipy/pull/2092

The PR:
- updates the fetch so the INSTALLATION.md, CONTRIBUTING.md, and CODE_OF_CONDUCT.md are copied.
- turns the installation page, the contributing page and the code of conduct page into templates, so the content is directly injected from the files in taipy repository
- creates a new Step instantiated three times to re-use the content from taipy.

Note: The changes on gitignore will be reverted once merged. Indeed, the installation.md file must be removed from origin. Once it is removed, we can ignore it again.


FabienLelaquais (approved):
looks good!

PR #1193, opened 2024-11-04
Allows float values for gallery article order

AlexandreSajus:
Currently, articles in the gallery are sorted according to an integer value in the header of the article:
image

Since the values are integers, adding an article between two articles forces contributors to modify the order value of many articles: if I want to place an article at position 6, I must take all articles from position 6 to N and increment their order by 1.

By accepting float values for order, we allow contributors to place an article at position 6 by setting the order at 6.5

### DIFF
--- a/tools/_setup_generation/items/item.py
+++ b/tools/_setup_generation/items/item.py
@@ -18,7 +18,7 @@ class Item:
             self.category = self.header.get("category")
             self.data_keywords = self.header.get("data-keywords")
             self.short_description = self.header.get("short-description")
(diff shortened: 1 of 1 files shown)


FlorianJacta:
Did you build the doc with this too?

AlexandreSajus:
Yes, and it works. I added a demo in #1194 with `order: 11.5` and it placed it between the 11th and the 12th article

FlorianJacta (approved):
LGTM

jrobinAV:
Lovely!

Issue #1226, opened 2024-12-04
BUG- Missing documentation for Polars


FlorianJacta:
**Bug description**
Release notes mentionned that we now support Polars but I have no idea how to use it. I tried putting different strings in exposed type without success.

Either this feature does not exist at all and this is an error inside the doc or if it exists, there are no ways to know how to use it.

This is also what I get when I try to run a code using polars.

```
[2024-12-04 11:14:00.518][Taipy][ERROR] The `exposed_type` of DataNodeConfig `output_data` must be either "pandas", "numpy", or a custom type. Current value of property `exposed_type` is "polars.LazyFrame".
Configuration errors found. Please check the error log for more information.
```

Polars is not mentionned.

Test Code:

```python
from taipy import Config, Gui
import taipy.gui.builder as tgb
import taipy as tp


def handle_preprocessing(data):
    # some polars code to preprocess data

    return data


input_data_cfg = Config.configure_data_node(
    id="input_data",
    storage_type="csv",
    default_data={"a": [1, 2, 3], "b": [4, 5, 6]},
    exposed_type="polars.LazyFrame",
)

output_data_cfg = Config.configure_data_node(
    id="output_data",
    storage_type="csv",
    exposed_type="polars.LazyFrame",
)

scenario_cfg = Config.configure_scenario(
    id="scenario",
    task_configs=[
        Config.configure_task(
            id="task",
            input=input_data_cfg,
            output=output_data_cfg,
            function=handle_preprocessing,
        )
    ],
)

if __name__ == "__main__":
    tp.Orchestrator().run()
    scenario = tp.create_scenario(scenario_cfg)
    with tgb.Page() as page:
        tgb.scenario_selector("{scenario}")
        tgb.scenario("{scenario}")

        tgb.data_node(
            lambda scenario: scenario.output_data if scenario is not None else None
        )

        tgb.chart(
            lambda scenario: (
                scenario.output_data.read() if scenario is not None else None
            )
        )

    Gui(page).run()
```

jrobinAV:
Must be documented on both develop and 4.0.X versions.

FlorianJacta:
Should you be assigned to this and have this issue in review/Q&A? @toan-quach

jrobinAV:
@toan-quach @FlorianJacta Is this issue solved?

FlorianJacta:
Is there a PR on it?

toan-quach:
[...]

PR #1259, opened 2025-01-14
bug/#1226 added doc for polars

toan-quach:
#1226

### DIFF
--- a/docs/userman/scenario_features/data-integration/data-node-usage.md
+++ b/docs/userman/scenario_features/data-integration/data-node-usage.md
@@ -263,12 +263,11 @@ The following examples represent the results when reading from a CSV data node w
     === "exposed_type = "pandas""
 
         ```python
-        pandas.DataFrame
-        (
-                     date  nb_sales
-            0  12/24/2018      1550
-            1  12/25/2018      2315
-            2  12/26/2018      1832
+        pandas.DataFrame(
+            {
+                "date": ["12/24/2018", "12/25/2018", "12/26/2018"],
+                "nb_sales": [1550, 2315, 1832]
+            }
         )
         ```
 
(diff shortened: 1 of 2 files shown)


docs/userman/scenario_features/data-integration/data-node-usage.md
```diff
@@ -293,6 +294,27 @@ The following examples represent the results when reading from a CSV data node w
...
+        shape: (3, 2)
+        ┌────────────┬──────────┐
+        │ date       ┆ nb_sales │
+        │ ---        ┆ ---      │
+        │ str        ┆ i64      │
+        ╞════════════╪══════════╡
+        │ 12/24/2018 ┆ 1550     │
+        │ 12/25/2018 ┆ 2315     │
+        │ 12/26/2018 ┆ 1832     │
+        └────────────┴──────────┘
```
jrobinAV: This is not Python code. I wonder if having some actual Python code that builds an equivalent Polars Dataframe is better. We should also do the same for pandas. What do you think? 
```
polars.DataFrame({
    "date": ["12/24/2018", "12/25/2018", "12/26/2018"]
    "nb_sales": [1550, 2315, 1832]
})
```

toan-quach: oh? I thought it's a representation of the expected output of polars Dataframe, sure I can replace it with actual polars code :D

jrobinAV: Yes, today, it's indeed a representation of the expected output. 
https://docs.taipy.io/en/develop/userman/scenario_features/data-integration/data-node-usage/#__tabbed_1_1
Would a real Python code make the documentation more clear? That's my question, but it's discussable.

Maybe @FlorianJacta and @AlexandreSajus have an opinion on that as well.

FlorianJacta: That's maybe a good idea

jrobinAV (commented):
Please align the pandas constructor Python code as well. They are incorrect.

Thx

jrobinAV (approved):
Neat!

PR #1260, opened 2025-01-17
Fix - Update dev Dockerfile example to optimize caching

trgiangdo:
Update the Dockerfile example to follow up [Florian's comment](https://github.com/Avaiga/taipy-enterprise/pull/609#discussion_r1918689446)


FlorianJacta (approved):
LGTM

Issue #911, opened 2024-04-10
Process links from the doc captured in Json


FabienLelaquais:
**Description**

taipy/gui/viselements.json, in particular, holds links to parts of the user man (or it could be the Reference Manual).
We need to make sure that in all contexts, links in the documentation texts are processed in a way that makes sense in the context.

Targets are:
- Reference Manual for visual elements (/refmans/gui/viselements/generic/<element_name>/)
- Page Builder API (/refmans/reference/pkg_taipy/pkg_gui/pkg_builder/<element_name>/)
- Autocompletion in IDEs (pyi generation)

**Expected change**
Links are active when possible, driving to the proper target documentation page.

PR #1291, opened 2025-03-10
Fix element cross-refs in all contexts

FabienLelaquais:
Addresses #911.

### DIFF
--- a/tools/check_site.py
+++ b/tools/check_site.py
@@ -1,48 +1,164 @@
+# ################################################################################
+# check_site.py
+#   Checks all links (local and external) in all files of a generated web site.
+# ################################################################################
+
 import os
 import re
-import logging
+import requests
+import threading
 
-SITE_DIR = "site"
+all_local_files = {}
+external_links = {}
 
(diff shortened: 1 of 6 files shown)


tools/check_site.py
```diff
@@ -1,48 +1,164 @@
...
+            if threads is not None and href not in external_links:
+                external_links[href] = ""
+                thread = threading.Thread(target=is_valid_url, args=(href, file_path))
+                thread.start()
+                threads.append(thread)
+
+        elif (
+            href.startswith("javascript:") or href.startswith("mailto:") or href.startswith("#")
+        ):  # Scrip, mail or internal ref link
+            ...
```
FredLL-Avaiga: pass ?

FredLL-Avaiga (approved):

tools/check_site.py
FabienLelaquais: Changed (and at another place as well). Thx

FredLL-Avaiga (approved):

PR #1293, opened 2025-03-11
Explicit supported Python versions.

jrobinAV:
SInce we do not support 3.13 yet, this PR explicitly specifies which Python version is supported.


FabienLelaquais (changes requested):
Please drop - just should be generated - new issue please.

jrobinAV:
https://github.com/Avaiga/taipy-doc/issues/1296

PR #1310, opened 2025-04-22
Change gallery section

FlorianJacta:
Change Gallery Section to be more use case driven.

image

image

image

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -119,7 +119,7 @@ mkdocs.yml
 
 /docs/gallery/index.md
 /docs/gallery/finance/index.md
-/docs/gallery/decision_support/index.md
+/docs/gallery/manufacturing_logistics/index.md
 /docs/gallery/llm/index.md
 /docs/gallery/visualization/index.md
 /docs/gallery/other/index.md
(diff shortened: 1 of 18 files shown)


FlorianJacta:
Should I try to merge it now? @jrobinAV

jrobinAV:
Yes, let's try.

FlorianJacta:
I removed the Draft tag of this PR

.gitignore
```diff
@@ -150,3 +150,4 @@ mkdocs.yml
 !/docs/userman/gui/gui_example.ipynb
 
 /docs/xrefs
+/taipy.clone
```
jrobinAV, line 153: What is that?

jrobinAV (approved):

.gitignore
FlorianJacta: I think this was when I fetched Taipy. I had this folder appear. Do you want me to drop this?

jrobinAV: No. We don't want it to be committed. If it is generated in your environment, we can keep the line in the .gitignore.

PR #1320, opened 2025-05-27
Prepare release notes for 4.1

jrobinAV:
Prepare release notes for taipy, core and enterprise packages.
Must be updated with actual links later when refman is ready.


FabienLelaquais (approved):
👍

PR #1323, opened 2025-05-28
Fix import in code snippet

### LINKED ISSUE #1322: BUG-Error in a code snippet
There is an error here: https://docs.taipy.io/en/latest/userman/gui/pages/navigate/#using-the-navigate-function

There is no *navigate* in taipy.

```python
from taipy import Gui, navigate
```

->

```python
from taipy.gui import Gui, navigate
```

FlorianJacta:
Resolves: #1322


FabienLelaquais (approved):
👍

PR #1341, opened 2025-07-09
Homogeneity of the parameters in on change callback function examples.

FabienLelaquais:
As @arcanaxion mentioned, parameter names are not consistent across the documentation.


FabienLelaquais:
Wrong origin branch!

PR #1342, opened 2025-07-09
Homogeneity of the parameters in on change callback function examples.

FabienLelaquais:
As @arcanaxion mentioned, parameter names are not consistent across the documentation.

Note that needs to be performed on the develop branch when merged.

### DIFF
--- a/docs/refmans/gui/viselements/generic/charts/advanced.md_template
+++ b/docs/refmans/gui/viselements/generic/charts/advanced.md_template
@@ -442,9 +442,9 @@ selected_indices = []
 
 mean_value = 0.
 
-def on_change(state, var, val):
-    if var == "selected_indices":
-        state.mean_value = numpy.mean([data["y"][idx] for idx in val]) if len(val) else 0
+def on_change(state, var_name, var_value):
+    if var_name == "selected_indices":
+        state.mean_value = numpy.mean([data["y"][idx] for idx in var_value]) if len(var_value) else 0
 ```
 
 The part you should be looking at is the use of the default callback property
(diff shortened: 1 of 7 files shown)


docs/refmans/gui/viselements/generic/date_range.md_template
```diff
@@ -52,8 +52,8 @@ The date range selector looks like this:
 
 An `on_change` callback can be set to retrieve the user's selection:
 ```python
-def on_change(s: State, name: str, value: any):
-    if name == "dates":
+def on_change(s: State, var_name: str, var_value: any):
```
arcanaxion: Minor comment but might as well change `s` to `state` in this PR. I think this is the only occurrence as a callback function parameter that we abbreviate `state` as `s`. I prefer abbreviating it only in lambdas and in a context manager

arcanaxion (approved):
😃

docs/refmans/gui/viselements/generic/charts/advanced.md_template
```diff
@@ -442,9 +442,9 @@ selected_indices = []
 
 mean_value = 0.
 
-def on_change(state, var, val):
-    if var == "selected_indices":
-        state.mean_value = numpy.mean([data["y"][idx] for idx in val]) if len(val) else 0
+def on_change(state, var_name, var_value):
```
jrobinAV, line 445: We can use type hinting as much as possible in our examples, thereby spreading a good practice.

jrobinAV (approved):
Optionally, we can use this PR to start showing type hinting in our examples.

arcanaxion (approved):

PR #1343, opened 2025-07-09
Doc/1336/remove quotes tree lambda

### LINKED ISSUE #1336: BUG-Error in how adapter uses lambda expression
**Bug description**
The documentation about the tree elements has an error in how it displays lambda expressions: it shows them between quotes "", but it shouldn't.

Here's the link: https://docs.taipy.io/en/release-4.1/refmans/gui/viselements/generic/tree/#__tabbed_7_3

This is the wrong code:

```python
import taipy.gui.builder as tgb
...
tgb.tree("{user_sel}", lov="{users}", type="User", adapter="lambda u: (u.id, u.name, u.children)")
```

--> `adapter` should be rewritten: `adapter=lambda u: (u.id, u.name, u.children)` without quotes.

**How I'd fix it**
In the Markdown doc, this is referenced as such:

```md
!!! taipy-element
    default={user_sel}
    lov={users}
    type=User
    adapter=lambda u: (u.id, u.name, u.children)
```

I think it should be written like this:

```
adapter:f=lambda u: (u.id, u.name, u.children)
```

**I'd be happy to take this issue if you agree :)**

enarroied:
Hi!

This commit has 3 changes:
* I modified the pre-commit-onfig file. I don't have a GitLab account, + I don't use Python 3.9. Maybe you can discard that part of the PR? Let me know if I did something very wrong here. Otherwise, all tests passed.
* I changed a line in the Pipfile, which was missing "==" and gave me errors when building the environments. It's 100% unrelated to the issue. Do you want to track it in a separate issue?
* I changed the line in tree.md, as described in the issue, so it Fixes #1336


enarroied:
hi! This is the first time I contribute to any of your projects (with a PR), I think I read you want it to be left as draft, do you want me to mark it as ready for review? Thanks!

.pre-commit-config.yaml
```diff
@@ -33,8 +33,8 @@ repos:
     hooks:
     - id: black
       args: [--line-length=120]
-      language_version: python3.9
--   repo: https://gitlab.com/pycqa/flake8
+      language_version: python3.12
```
FabienLelaquais: We are trying to ensure every thing builds and runs smoothly with Python 3.9.
Introducing 3.12 at this step does not sound like the best idea. Why do that?

FabienLelaquais (changes requested):
Quick question - the rest is just fine.

.pre-commit-config.yaml
enarroied: Hi, 
This is my first contribution. I currently use Python 3.12 for many projects.
I'll try to be mindful of this next time.
Do I remove this now, or do I wait for the second review?

jrobinAV: Please proceed, I agree with Fabien and have no further remarks.

enarroied: I changed my env to 3.9 and changed that line :)

arcanaxion:
@FabienLelaquais Ready for your review

FabienLelaquais (approved):
Looking good!

.pre-commit-config.yaml
FabienLelaquais: Thanks!

PR #1347, opened 2025-10-21
Use the same parameters names for all on_change callbacks across the documentation

FabienLelaquais:
- on_change signature homogeneity
- fix parameter names appearing in the RefMan pages TOC section.


jrobinAV (approved):
Neat

PR #1361, opened 2026-03-13
Suppress warning in documentation generation

FabienLelaquais:
When generating the Python API generation, griffe and mkdocs issue warnings that are annoying, confusing, and useless.


jrobinAV (approved):
Oh yeah !

